PR #5, opened 2019-02-14
added user not registered email and env variable for template id

joepike:
When user has been invited but has not yet registered they will receive an email notifying them to register for an eGAR account before they can login.

**Definition of Done**


src/app/user/login/post.controller.js
```diff
@@ -38,7 +38,19 @@ module.exports = (req, res) => {
           if (Object.prototype.hasOwnProperty.call(user, 'message')) {
             logger.info(`Invalid email entered: ${usrname}`);
             cookie.setUserVerified(false);
-            res.redirect('/login/authenticate');
+            console.log(config.NOTIFY_NOT_REGISTERED_TEMPLATE_ID);
```
Meepit: Remove or change to logger statement

src/app/user/login/post.controller.js
```diff
@@ -38,7 +38,18 @@ module.exports = (req, res) => {
           if (Object.prototype.hasOwnProperty.call(user, 'message')) {
             logger.info(`Invalid email entered: ${usrname}`);
             cookie.setUserVerified(false);
-            res.redirect('/login/authenticate');
+            emailService
```
Meepit: Probably worth checking to see if the message == user not registered
With the above an unverified user will get the email which shouldn't happen

PR #7, opened 2019-02-14
Create local dev environment

pawniemiec:
Local dev environment
Please read README.md for details

**Definition of Done**
- [X] Have you built the code locally and confirmed its working?
- [X] Is the build confirmed to be passing on CI?
N/A - no need to add any tests as there is no functionality change
N/A - no need to add any tests as there is no functionality change
- [X] Have you updated any relevant documentation as part of this change?
- [X] Has this change been tested against the Acceptance Criteria?
N/A - only dev environment

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,100 @@
 # Submit a GAR Public Site UI
 
-## Docker Usage
-The following depends on a .env file in the root of this project with the required environment variables. This will start a local test instance of keycloak for local testing.
+## Requirements
+- docker
+- docker-compose
+- node
+- npm
+- your email added to GOV Notify Dev whitelist by repository administrator
 
+## Development
+
+The following depends on a `.env.dev` file in the root of this project with
+the required environment variables. To obtain it please contact repository administrator.
+
+This will start local instances of:
+- Database (used for both Frontend and Backend)
+- API (you need a local copy of `api` docker image loaded, 
+       please contact repository administrators to obtain it)
+- ClamAV engine
+- ClamAV REST API 
+- Frontend
+
+Please note - due to security configuration, the following lines need to be 
+commented out for local development, but not committed back to the repository:
 ```
-docker-compose down -t 1
-docker-compose build
-docker-compose up -d
+./src/server.js (lines 95-98):
+  // app.use(csrf({cookie: {
+  //   httpOnly: true,
+  //   secure: true
+  // }}));
+
+./src/server.js (lines 103-106):
+    // var token = req.csrfToken();
+    // res.cookie('XSRF-TOKEN', token);
+    //
+    // res.locals.csrfToken = token;
 ```
 
-These steps will create 4 endpoints:
-- https://127.0.0.1:10443 -> main site secured with unverified TLS certificate (ignore CERT INVALID errors)
-- http://127.0.0.1:3000 -> node site
-- http://127.0.0.1:3310 -> ClamAV endpoint
 
+To start clean:
+```
+docker-compose up --build -d
+```
+These steps will create 5 endpoints:
+- http://localhost:5432 -> Database
+- http://localhost:5000 -> API
+- http://localhost:3310 -> ClamAV engine
+- http://localhost:8080 -> ClamAV REST API
+- http://localhost:3000 -> Node site
+
+
+Once it's up, you need to:
+- whitelist yourself in database:
+```
+for line in $(egrep -v "^#|^$" .env.dev); do export $line; done
+docker exec -ti database sh -c "PGPASSWORD=${POSTGRES_PASSWORD} psql -U ${POSTGRES_USER} ${POSTGRES_DB}"
+```
+
+Once inside database REPL run these commands.
+Don't forget to update your email address in the INSERT command:
+```
+CREATE TABLE "session" (
+  "sid" varchar NOT NULL COLLATE "default",
+	"sess" json NOT NULL,
+	"expire" timestamp(6) NOT NULL
+)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,19 +1,87 @@
...
+## Requirements
+- docker
+- docker-compose
+- node
+- npm
+- your email added to GOV Notify Dev whitelist by repository administrator
+
+## Development
+
+The following depends on a `.env.dev` file in the root of this project with
```
SvetlaVerma: Something worth mentioning is sourcing .env.dev won't work in the mac environment as it doesn't have the export command as needed by mac. Unless the administrator gives it with export command.

docker-compose.yml
```diff
@@ -2,26 +2,49 @@ version: '3'
...
-     - POSTGRES_DB=db
+      - CLAMD_HOST=clamav
+    depends_on:
+      - clamav
+    ports:
+      - "8080:8080"
+
+  node:
+    build: .
+    container_name: node
```
SvetlaVerma: The docker-compose works fine and I can see all 5 docker containers up locally. However I don't get the registration email in my inbox, so might be missing some step here.

README.md
pawniemiec: I have replaced the `source` command with `for` loop, covering both MacOS and POSIX Linux

docker-compose.yml
pawniemiec: I cannot replicate locally - need to cooperate with you on that.

README.md
```diff
@@ -1,19 +1,77 @@
...
+- http://localhost:5000 -> API
+- http://localhost:3310 -> ClamAV engine
+- http://localhost:8080 -> ClamAV REST API
+- http://localhost:3000 -> Node site
+
+
+Once it's up, you need to:
+- whitelist yourself in database:
+```
+docker exec -ti database sh -c "PGPASSWORD=${POSTGRES_PASSWORD} psql -U ${POSTGRES_USER} ${POSTGRES_DB}"
```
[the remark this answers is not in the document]
nigelhughes552: Coming from a Windows background I wasn't exactly sure what was going on (no egrep with Windows!) but got gist of what it was doing. A quick scan of [https://docs.docker.com/engine/examples/postgresql_service/](url) gave me a bit more info. But then again not being that UNIX savvy wasn't sure what psql was. I found that by installing PostgreSQL I could run the psql login command manually.

README.md
```diff
@@ -1,19 +1,96 @@
...
+To start stopped containers:
+```
+docker-compose start
+```
 
+To apply your changes to Frontend only:
+```
+docker-compose stop node
+docker-compose up --build -d node
+```
```
SvetlaVerma: Looks good Pawel, the only comment I have now is we should also have a command in README for running the tests locally. For a developer trying to run the local environment for a story, they will definitely try to run the tests so the command in README will help.

nigelhughes552: Re Non Docker Usage

Please excuse my NOOB understanding of the system but if you don't use the Docker method wouldn't you need a local database etc available?

pawniemiec: That's correct @nigelhughes552. 
Non docker/docker-compose setup would require you to install database, compile API code and start it, install Clamav service, install nginx proxy and configure it to work with clamav and have node/npm.
Quite a lot of work comparing to 1 `docker-compose up --build - d` command 😉

README.md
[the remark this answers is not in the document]
pawniemiec: @nigelhughes552 
So you need to export all environment variables to your local environment.
This will pass values to the `docker exec` command.
`docker exec` -> execute command in side docker 
`-ti database` -> do it in running image called `database` in interactive mode
`sh -c "PGPASSWORD=${POSTGRES_PASSWORD} psql -U ${POSTGRES_USER} ${POSTGRES_DB}"` -> command to execute (effectively connect to database using environment variables

`psql` command is executed inside running docker command, hence there is no need to have it installed locally (though as the Postgres port is exposed to localhost, you are able to connect to database via localhost:5432 address:port using any possible drivers)

README.md
```diff
@@ -1,19 +1,77 @@
...
+- API (you need a local copy of `api` docker image loaded, 
+       please contact repository administrators to obtain it)
+- ClamAV engine
+- ClamAV REST API 
+- Frontend
+
+Please note - due to security configuration, the following lines need to be 
+commented out for local development, but not committed back to the repository:
+```
+./src/server.js (lines 95-98):
```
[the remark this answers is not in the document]
pawniemiec: The fix would require:
- reverse proxy running
- an external certificate to be installed inside reverse proxy
It would involve some work and associated cost. There is no time and budget for that in Sprint 12 unfortunately

docker-compose.yml
pawniemiec: Issue solved. Docker image running had time skew. 
Docker service restart fixed the time skew issue, registration and authentication emails are now being received.

pawniemiec:
@SvetlaVerma & @nigelhughes552 - It seems I cannot resolve discussions.
Can you please resolve them, so I can merge this branch?

README.md
```diff
@@ -1,19 +1,77 @@
...
+- node
+- npm
+- your email added to GOV Notify Dev whitelist by repository administrator
+
+## Development
+
+The following depends on a `.env.dev` file in the root of this project with
+the required environment variables. To obtain it please contact repository administrator.
+
+This will start a local instances of:
```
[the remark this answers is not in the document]
pawniemiec: Good catch, removed.

PR #13, opened 2019-02-18
updated gar craft page styling

joepike:
Updates the garfile page for adding aircraft to match the prototype, with the exception that radio buttons for add saved craft are used instead of checkboxes.

**Definition of Done**

### DIFF
--- a/src/app/garfile/craft/index.njk
+++ b/src/app/garfile/craft/index.njk
@@ -3,137 +3,127 @@
 {% import "common/templates/includes/macros.njk" as m %}
 
 {% block pageTitle %}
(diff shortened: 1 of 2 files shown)


src/app/garfile/craft/index.njk
```diff
@@ -3,137 +3,127 @@
 {% import "common/templates/includes/macros.njk" as m %}
 
 {% block pageTitle %}
-	Create GAR | Arrival
+	Add Aircraft to GAR
```
pawniemiec: For consistency - can the title be `Create GAR | Aircraft` please?

PR #16, opened 2019-02-19
EGAR-441 Sign In Page - Styling review

nigelhughes552:

**Definition of Done**

### DIFF
--- a/src/app/user/login/index.njk
+++ b/src/app/user/login/index.njk
@@ -19,7 +19,6 @@
 
 {% block content %}
 
-
   {% import "common/templates/includes/macros.njk" as m %}
 
   {% if errors %}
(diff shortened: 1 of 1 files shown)


src/app/user/login/index.njk
```diff
@@ -47,9 +46,9 @@
 			<p class="govuk-body"><a href="/user/register">Create an account</a></p>
 
 			<button type="submit" class="govuk-button">
-				Agree and submit
+				Sign in
```
Meepit: Sign in button should be above the create an account section according to screenshot / prototype

Meepit:
The `Need to register an account` heading should be larger (govuk-heading-m)

src/app/user/login/index.njk
nigelhughes552: That was a big miss!- Must have gone screen blind. Because my commits weren't being signed I had to squash my commits and force push (with the help of @pawniemiec ). So hopefully this and the heading styling should be fixed.

PR #27, opened 2019-02-22
Only create GAR on file upload

Meepit:
Correct back button href

Update arrival lat/long on review page

Clicking back on the departure page will take the user to the gar creation page rather than file upload (if user uploaded via file) 

The GAR will only be created after file upload to prevent unexpected manifest behaviour 

Update review page to show correct arrival lat/long values.

**Definition of Done**


src/app/api/uploadgar/index.js
```diff
@@ -117,20 +118,34 @@ router.post('/uploadgar', upload.single('file'), (req, res, data) => {
...
+
+        Promise.all([crewUpdate, passengerUpdate, voyageUpdate])
+          .then(() => {
+            logger.info('Updated GAR with excel data');
+            return req.session.save(() => res.redirect('/garfile/departure'));
+          })
+          .catch((err) => {
+            logger.error('Failed to update API with GAR information');
+            logger.error(err);
+            req.session.failureMsg = 'Failed to read GAR';
```
SvetlaVerma: The logged error and the error message seem different? The logged error is "failed to update API with GAR information and the req.session.failureMsg we are setting is "Failed to read GAR".

src/app/api/uploadgar/index.js
```diff
@@ -117,20 +118,34 @@ router.post('/uploadgar', upload.single('file'), (req, res, data) => {
...
+            req.session.failureIdentifier = 'file';
+            return res.redirect('garfile/garupload');
```
SvetlaVerma: As a change in functionality, specially because we fixed arrival long and lat, do we need to fix any e2e tests as a result of this? (I do understand though I wont be able to see it as e2e tests repo are not opensourced).

Meepit: In my opinion, no end to end tests need to change as a result of this. E2E tests are testing integration between the various systems, this change only affects the view.

PR #33, opened 2019-02-25
Whitelist

joepike:
Adds conditional check for WHITELIST_REQUIRED env variable flag so that the email is only checked in the whitelist for production and development

**Definition of Done**

### DIFF
--- a/src/app/user/register/post.controller.js
+++ b/src/app/user/register/post.controller.js
@@ -8,6 +8,7 @@ const sendTokenService = require('../../../common/services/send-token');
 const userCreateApi = require('../../../common/services/createUserApi');
 const tokenApi = require('../../../common/services/tokenApi');
 const whitelist = require('../../../common/services/whiteList');
+const config = require('../../../common/config');
 
 module.exports = (req, res) => {
   logger.debug('In user / register post controller');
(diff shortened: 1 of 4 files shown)


src/app/user/register/post.controller.js
```diff
@@ -56,54 +57,62 @@ module.exports = (req, res) => {
...
+      if (config.WHITELIST_REQUIRED) {
+        logger.info('Starting whitelist check');
+        whitelist.isWhitelisted(usrname)
+          .then((result) => {
+            if (result) {
+              createUser();
+            } else {
+              // Not whitelisted
+              cookie.setUserEmail(null);
+              res.redirect('/user/regmsg');
```
Meepit: save session before redirect

src/app/user/register/post.controller.js
```diff
@@ -56,54 +57,62 @@ module.exports = (req, res) => {
...
+    function createUser() {
+      logger.info('Creating the user in the db');
+      userCreateApi.post(fname, lname, usrname, cookie.getInviteUserToken())
+        .then((dbUser) => {
+          if (Object.prototype.hasOwnProperty.call(JSON.parse(dbUser), 'message')) {
+            logger.info('Failed to register user in db');
+            logger.info(`${JSON.parse(dbUser).message}`);
             cookie.setUserEmail(null);
-            res.redirect('/user/regmsg');
+            return res.redirect('/user/regmsg');
```
Meepit: Save session before redirect

src/app/user/register/post.controller.js
```diff
@@ -56,54 +57,62 @@ module.exports = (req, res) => {
...
-                  });
-              })
-              .catch((err) => {
-                logger.error(`Failed to create ${usrname} in DB`);
-                logger.error(err);
-                res.render('app/user/register/index', { cookie, errors: [regFailureError] });
-              });
-          } else {
-            // Not whitelisted
+      if (config.WHITELIST_REQUIRED) {
```
Meepit: You need to cast the env var from string to bool at some point. Otherwise you're checking if('true') and if('false') rather than if(true) and if(false)

src/app/user/register/post.controller.js
```diff
@@ -46,6 +47,8 @@ module.exports = (req, res) => {
 
   const regFailureError = { message: 'Registration failed, try again' };
 
+  const isWhitelistRequired = (config.WHITELIST_REQUIRED == 'true');
```
Meepit: call .toLowerCase() on config.WHITELIST_REQUIRED so it isn't case sensitive

PR #47, opened 2019-03-10
Nojira cookie policy

Meepit:

**Definition of Done**


src/common/templates/base-template.njk
```diff
@@ -43,6 +45,12 @@
       }) }}
     {% endblock %}
 
+    {% block cookie_message %}
+    <div class="app-cookie-banner" style="display: none;">
+      <p class="app-cookie-banner__message govuk-body-s">GOV.UK uses cookies to make the site simpler. <a href="/cookies">Find out more about cookies</a></p>
```
SvetlaVerma: It says GOV.uk here. Should it be "Submit a GAR" ?

PR #57, opened 2019-03-14
Remove Private Beta Production URL

pawniemiec:
Remove Private Beta Production URL
Remove Service test URL from SIT and Staging


Meepit:
Looks good to me

PR #58, opened 2019-04-07
Add validation for same arrival / departure ports

Meepit:

**Definition of Done**

### DIFF
--- a/src/common/utils/validator.js
+++ b/src/common/utils/validator.js
@@ -264,6 +264,18 @@ function validateChains(chains) {
 }
 
 /**
+ * false if all values in array are equal, else true
+ * @param {Array} values
+ * @returns {Bool}
+ */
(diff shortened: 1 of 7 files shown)


src/common/utils/validator.js
```diff
@@ -263,6 +263,18 @@ function validateChains(chains) {
...
+/**
+ * false if all values in array are equal, else true
+ * @param {Array} values
+ * @returns {Bool}
+ */
+function notSameValues(values) {
+  if(values.includes(null) || values.includes(undefined)) {
+    return true;
+  }
+  return !values.every(v => v.trim().toLowerCase() === values[0].trim().toLowerCase());
```
SvetlaVerma: As I said in the slack message, we should make an exception for zzzz. Will wait to see what does Adam say on this.

PR #63, opened 2019-05-20
update deprecated ingress annotation for auto-tls certs

rhysemmas:
remove deprecated annotation and include new one to ensure let's encrypt certificates are automatically issued and renewed by the cert-manager service running in kubernetes


Meepit:
Looks good to me

PR #70, opened 2019-06-04
EGAR-742 disable input field autocomplete

chanc789:
Essentially sets the "false" property for "autocomplete" to "off". This is in line with standards, but opens a whole kettle of fish regarding browsers and how they handle autocomplete/autofill in general, where the use of "false" was a happy coincidence within Chrome, which simply disabled it as long as it was not "on" or "off".

There are suggestions on line regarding the use of other properties like "new-password" which may be another possible avenue, and so this pull request might be the start of a series.

Alternative extremes are munging input field names and/or URLs in order to "trick" the browser into believing it is a different page, which comes at an obvious cost of potential redesign.

This is locally tested on Chrome and Safari.

**Definition of Done**

### DIFF
--- a/src/app/aircraft/add/index.njk
+++ b/src/app/aircraft/add/index.njk
@@ -37,7 +37,7 @@
     {% include "common/templates/includes/errors.njk" %}
   {% endif %}
 
-      <form action="/aircraft/add" autocomplete="false" class="form-group"
+      <form action="/aircraft/add" autocomplete="off" class="form-group"
       method="post" role="form">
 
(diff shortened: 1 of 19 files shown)


pawniemiec:
Should you wondering why there is `edited by pawniemiec` in the title of the description, I have accidentally clicked on one of the check-boxes, so I unselected it. 
BTW - I'm not in position to review this PR at the moment.

chanc789:
Closing, superseded by branch "egar-742-jquery-disable-autofill".

PR #72, opened 2019-06-11
EGAR-739 cookie hardening

nguyendhod:
Certain cookies in the app were not marked with http only or secure when they should be. This PR should address this, whilst setting secure when actually possible.

This has also addressed an issue with the need to disable CSRF tokens when locally developing, although this needs checking within the staging environment.

Logs should now prepend the file name in order to aid debugging in future.

**Definition of Done**


nguyendhod:
Added a couple of new commits which may potentially send the test coverage results to SonarQube, so needs another look at.

nguyendhod:
All green to merge

PR #73, opened 2019-06-12
Cast country codes and port codes to upper case

Meepit:

Grammar fixes, cast country and port codes to upper case

**Definition of Done**

### DIFF
--- a/src/app/aircraft/add/index.njk
+++ b/src/app/aircraft/add/index.njk
@@ -79,7 +79,7 @@
              <label class="govuk-label" for="craftbase">Aircraft home port / location</label>
           {{ m.error_message('craftbase', errors) }}
           <span class="govuk-hint">
-            Provide the IATA / ICAO code or coordinates (decimal). You are only required to enter coordinates if your port does not have an IATA/ICAO code. Provide coordinates to 4 decimal places.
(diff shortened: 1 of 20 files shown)


nguyendhod:
Pipe probably sufficient, but belts and braces:

1. _style="text-transform: uppercase;"_ will be a CSS based approach to make inputs uppercase.
2. post.controller.js file could also uppercase just in case?

Otherwise the country and port codes can still be any random casing as per user input in the db. Not critical I suppose...

src/app/aircraft/add/index.njk
```diff
@@ -79,7 +79,7 @@
              <label class="govuk-label" for="craftbase">Aircraft home port / location</label>
           {{ m.error_message('craftbase', errors) }}
           <span class="govuk-hint">
-            Provide the IATA / ICAO code or coordinates (decimal). You are only required to enter coordinates if your port does not have an IATA/ICAO code. Provide coordinates to 4 decimal places.
```
SvetlaVerma: Verified locally and all looks good James. The only comment I don't find implemented is David suggestion for secition 3 of 8, that once an airport is added to the GAR, the following sentence should really disappear as it is no longer relevant for this GAR: "Add the details of a new aircraft you have not already saved". I still see this sentence once I have added an aircraft. can you check please?

Meepit:
Implemented uppercasing in postcontroller for consistency too. 
Still require upper casing in view because of historical data 

Also remove "Add a new aircraft you have not already saved" line when a registration is also present for the given GAR

PR #93, opened 2019-07-17
EGAR-764 empty table labels

nguyendhod:
When tables have no information, a message saying there are no records is displayed as a row in the empty table. The alternative is to hide the table completely and only display the label. This will depend on what business/testers think looks best.

Also sneaking in a little bit of EGAR-772 which is pagination for the aircraft page. This is missing E2E tests and unit tests, but is at least demonstrable and hopefully can provide some early feedback from staging.

**Definition of Done**


src/app/garfile/manifest/index.njk
```diff
@@ -33,25 +33,32 @@
 
 			<h2 class="govuk-heading-l">{{ __('nav_bar_people') }}</h2>
 				<p class="govuk-body-l">Add the details of people you have previously added</p>
-
+  </div>
+  <div class="govuk-grid-column-full">
+        {% if cookie.getSavedCraft().items | length %}
```
chanc789: Manifest page doesn't seems to need to check for 'savedCraft'.

src/common/templates/check-your-answers.njk
```diff
@@ -178,9 +179,10 @@
         {% endif %}
       </div>
     </dl>
-
+  </div>
     <div class="govuk-form-group {{ m.form_group_class('manifestTable', errors) }}">
     {{ m.error_message('manifestTable', errors) }}
+    <div class="govuk-grid-column-full">
```
chanc789: Use 'govuk-grid-column-two-thirds' to keep page format uniform.

src/common/templates/check-your-answers.njk
```diff
@@ -378,7 +387,8 @@
         {% endif %}
       </div>
     </dl>
-
+  </div>
+  <div class="govuk-grid-column-full">
```
chanc789: Use 'govuk-grid-column-two-thirds' to keep page format uniform.

PR #100, opened 2019-07-24
SGAR-746 ZZZZ ports

nguyendhod:
If an arrival or departure port is set to ZZZZ, it is expected that co-ordinates are entered. However, the application and possibly CBP as well will raise an error if the port codes are the same, despite in this case potentially being different co-ordinates.

This is an initial attempt to address is by referring to departure ports in this instance to 'YYYY', in the hopes that this should pass app validations and possibly CBP as well.

This pull request also folds in a fix for SGAR-765, which addresses an issue where the GAR aircraft and manifest pages do not behave as expected due to having more than one action button.

**Definition of Done**

### DIFF
--- a/src/app/garfile/arrival/index.njk
+++ b/src/app/garfile/arrival/index.njk
@@ -149,7 +149,7 @@
           <label class="govuk-label" for="arrivalLat">
             {{__('field_latitude')}}
           </label>
-          <span id="arrivalLattitude-hint" class="govuk-hint">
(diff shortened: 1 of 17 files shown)


chanc789:
Need to update the get controller as well. When reading from database 'YYYY' need to be displayed as 'ZZZZ'.

nguyendhod:
Still not sure whether this is the correct approach, as still unclear whether CBP would accept 'YYYY' as a port in this scenario. So in two minds about whether to flesh this out further at this stage...

PR #104, opened 2019-07-27
SGAR-715 and SGAR-946 unsubmitted gar caption and pre populated template caption

nguyendhod:
Couple of content changes, with refactoring of the tables in the dashboard screen and use of externalised strings in Departure and Arrival screens.

E2E tests are updated and PR raised, but these will pass without it (extra assertions rather changes of logic).

**Definition of Done**


nguyendhod:
I recall SGAR-715 being apparently of priority, so this could be merged in for quick test before sprint end. If so, then some E2E tests could be merged which begin testing organisations.

PR #107, opened 2019-07-31
Possible screen reader tweaks for login and registration

nguyendhod:
Removed the role button for the sign in, as it really is just a link rather a button to submit something. Leaving the button styling as it is the positive action for their respective screens.

Added aria-describedby attributes to hopefully aid input text inputs.

Refactored some variables, like 'Username' to 'username'.

Removed the unobtrusive validation attributes in the registration screen, as it is not used.

Externalised the strings for the login and registration screens.

This may address the login/registration screen reader issues. No way to prove or disprove.

Some refactoring around the html form input names, "Username" replaced with "username" just to begin some sense of consistency.

**Definition of Done**

### DIFF
--- a/src/app/home/index.njk
+++ b/src/app/home/index.njk
@@ -17,16 +17,17 @@
 {% block content %}
 <div class="govuk-grid-row">
 	<div class="govuk-grid-column-two-thirds">
-    {% if userSession.StatusChangedTimestamp%}
+    {% if userSession.StatusChangedTimestamp %}
       <h1 class="govuk-heading-m">
         <span class="heading-secondary govuk-hint">
-          You last logged in on {{userSession.StatusChangedTimestamp}}
+          {% set timestamp = userSession.StatusChangedTimestamp.toString() %}
+          {{__('last_logged_on', timestamp)}}
         </span>
-        Welcome back, {{cookie.getUserFirstName()}}
+        {{__('welcome_back_user', cookie.getUserFirstName())}}
       </h1>
     {% else %}
       <h1 class="govuk-heading-m">
-        Welcome, {{cookie.getUserFirstName()}}
+        {{__('welcome_user', cookie.getUserFirstName())}}
       </h1>
     {% endif %}
 	</div>
@@ -48,7 +49,7 @@
     <div class="govuk-grid-column-full">
       <h1 class="govuk-heading-xl">{{__('heading_create_gar')}}</h1>
       <p class="govuk-body-l">{{__('create_gar_caption_1')}}<a href="https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/392679/GAR_Template_with_logos_v_6_3_Jan_2015_LJ.xls">{{__('create_gar_caption_link')}}</a>{{__('create_gar_caption_2')}}</p>
-      <a id="start_now" class="govuk-button" href="/garfile/home" role="button">{{__('button_start_now')}}</a>
+      <a id="start_now" class="govuk-button" href="/garfile/home">{{__('button_start_now')}}</a>
     </div>
   </div>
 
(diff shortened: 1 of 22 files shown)


src/app/user/login/index.njk
```diff
@@ -44,21 +47,21 @@
 	<div class="govuk-grid-column-two-thirds">
 		<form action="login" autocomplete="off" method="POST">
 			<h1 class="govuk-heading-xl">{{ __('button_sign_in') }}</h1>
-			<div class="govuk-form-group {{ m.form_group_class('Username', errors) }}">
```
nguyendhod: Renamed "Username" to "username" as it was bugging me. Post controller then sets it as usrname when it's doing it's thing.

src/app/home/index.njk
```diff
@@ -17,16 +17,17 @@
 {% block content %}
 <div class="govuk-grid-row">
 	<div class="govuk-grid-column-two-thirds">
-    {% if userSession.StatusChangedTimestamp%}
+    {% if userSession.StatusChangedTimestamp %}
       <h1 class="govuk-heading-m">
         <span class="heading-secondary govuk-hint">
-          You last logged in on {{userSession.StatusChangedTimestamp}}
+          {% set timestamp = userSession.StatusChangedTimestamp.toString() %}
```
nguyendhod: i18n needed the timestamp as a String, hence the toString() call on what should be a Date object

src/app/user/register/index.njk
```diff
@@ -32,48 +31,48 @@
       <div class="govuk-form-group {{ m.form_group_class('userFname', errors) }}">
         <label class="govuk-label govuk-label--m" for="userFname">{{ __('field_given_name') }}</label>
         {{ m.error_message('userFname', errors) }}
-        <input autocomplete="off" class="govuk-input govuk-input--width-30"
-        data-val="true" data-val-required="The given name field is  required." id="userFname" name="userFname" style="width:100%"
```
nguyendhod: data-val stuff appears to be unobtrusive validation I think it's called, something jquery uses and stuff, or ASP.net.

I don't see jquery-validate imported into this app, and we know how the current validation works as it is, so this seems to be some hangover from another time.

src/app/user/register/index.njk
```diff
@@ -32,48 +31,48 @@
...
+            </a>
+          </li>
         </ul>
       </p>
 
       <div class="form-group">
         <input id="login" class="govuk-button" type="submit" value="Agree and submit" onclick="sendAnalytics(event, 'Register - submit', 'click')">
+        <input type="hidden" name="_csrf" value="{{ csrfToken }}" />
       </div>
-      <input type="hidden" name="_csrf" value="{{ csrfToken }}" />
```
nguyendhod: Not sure this makes a difference

src/app/verify/mfa/index.njk
```diff
@@ -30,29 +30,19 @@
...
+      <h1 class="govuk-heading-l">{{__('title_confirm_mfa_code')}}</h1>
 
-      <p class="govuk-body-l">
-        If the email that you entered is associated with an account then you will have received
-        an email with your login code. Please enter it below.
-      </p>
+      <p class="govuk-body-l">{{__('confirm_mfa_code_caption')}}</p>
 
       <form action="/login/authenticate" method="post">
-        <div class="govuk-form-group {{ m.form_group_class('mfa-authentication-code', errors) }}">
```
nguyendhod: Just changed from "mfa-authentication-code" to "mfa-code" just for the sake of less typing

src/server.js
```diff
@@ -162,7 +162,7 @@ function initialiseGlobalMiddleware(app) {
     // The below adds the csrfToken to the res.render function which should hopefully
     // allow for local development without this hack
     const _render = res.render;
-    res.render = function addToken(view, options, fn) {
+    res.render = function (view, options, fn) {
```
nguyendhod: ESLint complains that it is anonymous, and SonarQube complains that it should be anonymous, so deferred to SonarQube in this instance.

PR #109, opened 2019-07-31
SGAR-997 Departure and Arrival screen reader tweaks

nguyendhod:
Changes to improve screen reader support. Includes adding "Error: " prefixes to the title to allow a screen reader to inform a user earlier about an error.

More screen reader attributes to read out hints and errors for fields as applicable.

Added a possible improvement, where a user with a valid session now can go straight to the user dashboard when clicking "Start Now" instead of requiring to log in again. This should aid a little in QoL.

**Definition of Done**


nguyendhod:
Will do the other pull request first, tomorrow morning.

PR #112, opened 2019-08-01
SGAR-1008 cancel gar message not appearing

nguyendhod:
Original intention of this branch was to ensure the cancellation message appeared when cancelling a GAR. E2E tests which perform the cancellation of a GAR is perhaps the biggest value win from this.

This has now also lost focus and spread out a wide breadth. This includes:

Tweaks to the session TTL so it matches the 1 hour cookie time.

Experimental approach to changes to the page flow. On the Arrival and Departure screens, exiting now returns user to the GAR summary rather than the dashboard. If considered ok, will be used throughout the app.

Added ids to parts like the Further Information panels in the manifest screens to aid E2E test scripting.

**Definition of Done**


src/app/aircraft/edit/post.controller.js
```diff
@@ -34,13 +34,22 @@ module.exports = (req, res) => {
...
       craftApi.update(craftReg, craftType, craftBase, cookie.getUserDbId(), craftId)
         .then((apiResponse) => {
-          const parsedResponse = JSON.parse(apiResponse);
-          if (Object.prototype.hasOwnProperty.call(parsedResponse, 'message')) {
-            // API returned failure
-            res.render('app/aircraft/edit/index', { cookie, errors: [parsedResponse] });
-          } else {
-            // API returned successful
-            res.redirect('/aircraft');
+          try {
```
nguyendhod: This is a temporary solution to an issue where the app would return an unhelpful error if they attempted to rename an aircraft registration to one already in the DB.

This needs fixing properly by correcting the index in the DB to make user_id and registration or org_id and registration the unique indices.

src/app/garfile/arrival/index.njk
```diff
@@ -172,27 +172,11 @@
...
-            </a>
-          </p>
-          <p class="govuk-body-m">
-            <a href="/home">{{__('form_exit_without_saving')}}</a>
-          </p>
-        </div>
-        {# End buttons #}
-        <input type="hidden" name="_csrf" value="{{ csrfToken }}" />
-      </form>
+    {# Start buttons #}
```
nguyendhod: This is the result of using the macros for the potential new form submissions and exiting.

src/app/garfile/cancel/index.njk
```diff
@@ -25,17 +25,22 @@
...
   <br>
   <form action="/garfile/cancel" method="post">
-    <input class="govuk-button" type="submit" value="Cancel GAR">
+    <button id="cancel-gar" class="govuk-button" type="submit">{{__('title_cancel_gar')}}</button>
     <input type="hidden" name="_csrf" value="{{ csrfToken }}" />
   </form>
+  <form id="cancel-form" action="/garfile/view" method="post" role="form">
   <p class="govuk-body-m">
-    <a href="/home">Exit without cancelling</a>
+    <button type="submit" id="exit-without-cancelling" class="govuk-button govuk-button--secondary" data-module="govuk-button">
```
nguyendhod: Essentially a copy and paste of the revised exit button

src/app/garfile/cancel/index.njk
```diff
@@ -25,17 +25,22 @@
...
+    {{__('heading_cancel_gar')}}
   </h1>
   <p class="govuk-body-m">
-    You will need to ensure that a new GAR is submitted if this flight takes place.
+    {{__('cancel_gar_caption')}}
   </p>
   <br>
   <form action="/garfile/cancel" method="post">
-    <input class="govuk-button" type="submit" value="Cancel GAR">
+    <button id="cancel-gar" class="govuk-button" type="submit">{{__('title_cancel_gar')}}</button>
```
nguyendhod: WCAG suggests button is a little better than input with type="submit", so will need to double check if this is the case

src/app/garfile/cancel/post.controller.js
```diff
@@ -16,11 +16,15 @@ module.exports = (req, res) => {
       }).then(() => {
         req.session.successMsg = 'The GAR has been successfully cancelled';
         req.session.successHeader = 'Cancellation Confirmation';
-        res.redirect('/home');
+        req.session.save(() => {
```
nguyendhod: Not sure if this is necessary, but this definitely ensures the successMsg and successHeader get stored in the session before the redirect.

src/app/garfile/craft/index.njk
```diff
@@ -19,71 +19,71 @@
...
-{% endif %}
-</div>
-<div class="govuk-grid-column-two-thirds">
-  {% if cookie.getSavedCraft().items | length %}
-    <div class="govuk-form-group">
-      <input id="saveCraft" class="govuk-button" type="submit" name="buttonClicked" value="Add to GAR"
-      onclick="sendAnalytics(event, 'Add craft to GAR - submit', 'click')">
     </div>
-  {% endif %}
+    <div class="govuk-grid-column-full">
```
nguyendhod: I think we may at odds regarding indentations, as this has happened a lot in this pull request

src/app/garfile/craft/index.njk
```diff
@@ -131,8 +131,8 @@
         <a href="/home">{{__('form_exit_without_saving')}}</a>
       </p>
       <input type="hidden" name="_csrf" value="{{ csrfToken }}" />
-		</form>
-	</div>
+	  </div>
+  </form>
```
nguyendhod: Main tweaks are that the form is now one of the outer elements of the page, as it was actually badly done previously (a two thirds block had the start of the form inside it previously)

src/app/garfile/departure/index.njk
```diff
@@ -19,9 +19,6 @@
 
 {% block content %}
 
-
```
nguyendhod: Duplicate macro import so removed

src/app/garfile/departure/post.controller.js
```diff
@@ -85,7 +85,8 @@ const performAPICall = (cookie, res, buttonClicked) => {
       if (buttonClicked === 'Save and continue') {
         res.redirect('/garfile/arrival');
       } else {
-        res.redirect('/home');
```
nguyendhod: This is almost a hack of sorts.

If "Save and Exit" is pressed, we essentially want to redirect to the /garfile/view page, but the problem is a normal redirect will perform a GET, which doesn't work. It only accepts POST calls.

307 essentially translate to "POST again" with the same details, which works with the current /garfile/view and thus loads back the GAR for the summary display.

Would have been much simpler to have been a GET call or something, but essentially, any redirects previously to '/home' will need to be 307, '/garfile/view'

src/app/garfile/manifest/index.njk
```diff
@@ -26,19 +24,19 @@
...
-	<div class="govuk-grid-column-two-thirds">
-		<form action="/garfile/manifest" method="post">
-			<span class="govuk-caption-xl">{{__('section_count', 4, 8)}}</span>
-			<h1 class="govuk-heading-xl">{{__('heading_manifest_details')}}</h1>
-			<h2 class="govuk-heading-l">{{ __('title_people') }}</h2>
-			<p class="govuk-body-l">{{__('caption_manifest_details')}}</p>
-  </div>
-  <div class="govuk-grid-column-full">
-        {% if savedPeople | length %}
+  <form action="/garfile/manifest" method="post">
```
nguyendhod: Moved the form outside as it looked wrong having it inside a div that then is closed before the form.

src/app/garfile/manifest/index.njk
```diff
@@ -78,131 +76,131 @@
...
-        </table>
-  </div>
-  <div class="govuk-grid-column-two-thirds">
-    <button id="addToManifest" type="submit" name="buttonClicked" value="Add to GAR" class="govuk-button">Add to GAR</button>
-  </div>
-  <div class="govuk-grid-column-full">
-        <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
-  </div>
-  <div class="govuk-grid-column-two-thirds">
-				<h2 class="govuk-heading-l">New people</h2>
```
nguyendhod: Not sure what these large indentations were about...

src/app/garfile/manifest/index.njk
```diff
@@ -78,131 +76,131 @@
...
+            <td class="govuk-table__cell">
+              <button class="link btn-link" name="editPersonId" value="{{person.garPeopleId}}">Edit</button>
+            </td>
+          </tr>
+          <tr class="govuk-table__row">
+            <td class="govuk-table__cell" colspan="7">
+              <div>
+                <details role="group">
+                  <summary role="button" aria-controls="details-content-{{ loop.index }}" aria-expanded="false" title="{{__('further_information_for', person.firstName, person.lastName)}}">
+                    <span id="further-information-{{ loop.index }}" class="summary">{{__('label_manifest_further_personal_details')}}</span>
```
nguyendhod: Added these id's as the xpath in the E2E tests were pretty wordy and confusing

src/app/garfile/manifest/index.njk
```diff
@@ -78,131 +76,131 @@
...
+    </div>
+    <div class="govuk-grid-column-two-thirds">
+      <div class="govuk-form-group">
+        <label class="govuk-label" for="totalPersonsOnBoard">
+          {{__('total_persons_on_board')}}
+        </label>
 
-					<input class="govuk-input govuk-input--width-3" id="totalPersonsOnBoard" name="name" type="number" value={{manifest.items.length}}>
-				</div>
+        <input class="govuk-input govuk-input--width-3" id="totalPersonsOnBoard" name="name" readonly type="number" value="{{ manifest.items.length }}">
```
nguyendhod: Not sure why the display of the total is actually an input, maybe some screen reader thing?

It is set as readonly, because it kind of is, not sure if this affects accessibility.

src/app/garfile/view/post.controller.js
```diff
@@ -6,7 +6,10 @@ const garApi = require('../../../common/services/garApi');
 module.exports = (req, res) => {
   logger.debug('In garfile / view post controller');
   const cookie = new CookieModel(req);
-  const { garId } = req.body;
+  let { garId } = req.body;
```
nguyendhod: The GAR Summary page takes a POST call with "garId" expected to be in the form submitted.

Update now falls back to checking the session (via the cookie.getGarId()) if this is not set, so that the aforementioned 307 redirect does not require a POST submit with the GAR ID set, which would be derived from the cookie.getGarId() call anyway.

src/app/garfile/view/post.controller.js
```diff
@@ -6,7 +6,10 @@ const garApi = require('../../../common/services/garApi');
...
+  if (garId === undefined) {
+    garId = cookie.getGarId();
+  }
```
nguyendhod: Whether there should be additional checks in case both are undefined, depends.

src/app/garfile/view/post.controller.js
```diff
@@ -35,18 +38,17 @@ module.exports = (req, res) => {
         garpeople: parsedPeople,
         supportingdocs: supportingDocuments,
       };
+      renderContext.showChangeLinks = true;
       if ((parsedGar.status.name === 'Submitted') || parsedGar.status.name === 'Cancelled') {
-        logger.info('Rendering submitted GAR review page');
```
nguyendhod: Removed the submitted and un-submitted pages as there is not really any difference, nothing that the logic already could handle in a single page.

src/app/home/get.controller.js
```diff
@@ -11,8 +11,15 @@ module.exports = (req, res) => {
   const role = cookie.getUserRole();
   const orgId = cookie.getOrganisationId();
 
+  // Delete any GAR stored in the cookie session
+  cookie.session.gar = null;
+
   tokenApi.getLastLogin(cookie.getUserEmail())
     .then((userSession) => {
+      const { successHeader, successMsg } = req.session;
```
nguyendhod: GAR dashboard never had the logic in place to take the success message and display it, hence the point of this story.

src/app/home/index.njk
```diff
@@ -103,16 +107,16 @@
               <tr class="govuk-table__row">
                 <td class="govuk-table__cell">
                 {% if gar.registration === null %}
-                  <button class="govuk-link btn-link" name="garId" value="{{ gar.garId }}">{{__('value_aircraft_registration_null')}}</button>
+                  <button id="view-{{ name }}-{{ loop.index }}" class="govuk-link btn-link" name="garId" value="{{ gar.garId }}">{{__('value_aircraft_registration_null')}}</button>
```
nguyendhod: Items now have ids, more for simpler E2E test script authoring

src/common/models/Cookie.class.js
```diff
@@ -184,7 +184,7 @@ class Cookie {
   }
 
   setGarArrivalVoyage(voyageObj) {
-    if(!voyageObj.arrivalDate) {
+    if (!voyageObj.arrivalDate) {
```
nguyendhod: ESLint

src/common/models/Cookie.class.js
```diff
@@ -207,7 +207,7 @@ class Cookie {
   setGarDepartureVoyage(voyageObj) {
     if (!voyageObj.departureDate) {
       this.session.gar.voyageDeparture.departureDate = this.generateDate(voyageObj.departureDay,
-        voyageObj.departureMonth,voyageObj.departureYear);
+        voyageObj.departureMonth, voyageObj.departureYear);
```
nguyendhod: ESLint

src/common/templates/check-your-answers.njk
```diff
@@ -179,51 +179,51 @@
...
+              <td class="govuk-table__cell">{{person.documentNumber}}</td>
+              <td class="govuk-table__cell">{{person.nationality | upper}}</td>
+              <td class="govuk-table__cell">{{person.peopleType.name}}</td>
+            </tr>
+            <tr class="govuk-table__row">
+              <td class="govuk-table__cell" colspan="6">
+                <div>
+                  <details role="group">
+                    <summary role="button" aria-controls="details-content-{{ loop.index }}" aria-expanded="false" title="{{__('further_information_for', person.firstName, person.lastName)}}">
+                      <span id="further-information-{{ loop.index }}" class="summary btn-link">{{__('label_manifest_further_personal_details')}}</span>
```
nguyendhod: This was previously a convoluted xpath in the E2E test so an id simplifies things

src/common/templates/check-your-answers.njk
```diff
@@ -179,51 +179,51 @@
...
+              <td class="govuk-table__cell">{{person.dateOfBirth}}</td>
+              <td class="govuk-table__cell">{{person.documentNumber}}</td>
+              <td class="govuk-table__cell">{{person.nationality | upper}}</td>
+              <td class="govuk-table__cell">{{person.peopleType.name}}</td>
+            </tr>
+            <tr class="govuk-table__row">
+              <td class="govuk-table__cell" colspan="6">
+                <div>
+                  <details role="group">
+                    <summary role="button" aria-controls="details-content-{{ loop.index }}" aria-expanded="false" title="{{__('further_information_for', person.firstName, person.lastName)}}">
```
nguyendhod: This was previous "details-content-2" which seemed incorrect.

src/common/templates/includes/macros.njk
```diff
@@ -15,3 +15,29 @@
 		{% endif %}
 	{% endfor %}
 {% endmacro %}
+
+{% macro submit_buttons(analyticsPage, csrfToken, __) %}
```
nguyendhod: Bugs me a little that the i18n __ handle is a parameter. Probably lack of nunjucks knowledge there.

src/locales/en.json
```diff
@@ -17,7 +17,7 @@
     "welcome_tab_2_content_heading_1": "By email",
     "welcome_tab_2_content_1_block_1_1": "Send the ",
     "gar_form_link": "https://www.gov.uk/government/publications/general-aviation-operators-and-pilots-notification-of-flights",
-    "welcome_tab_2_content_1_block_1_2": "General Aviation Report form",
+    "welcome_tab_2_content_1_block_1_2": "General Aviation Report form (opens in a new window)",
```
nguyendhod: Accessibility articles online suggest that links that do things like open a new tab or window should have it noted for a screen reader to inform a user.

src/server.js
```diff
@@ -96,6 +96,7 @@ function initialisExpressSession(app) {
     genid: () => uuid(),
     store: new PgSession({
       conString: config.PUBLIC_SITE_DB_CONNSTR,
+      ttl: 60 * 60,
```
nguyendhod: Placed TTL of an hour here, to match the cookie maxAge. Will need to double check the effects.

nguyendhod:
Added comments throughout the code to try and help contextualise what has happened.

Unfortunately, this pull request is more a case of the start of bits of stories.

PR #113, opened 2019-08-03
SGAR-995 login registration screen reader

nguyendhod:

Essentially WAVE driven development. Errors were thrown when aria-describedby attribute contained the error reference when an error message was not present, so this is now determined before it is added using nunjucks.

WAVE also mentioned that phrases under the heading were possibly headings as well. This perhaps is a false positive, so up for debate whether they are changed to h2 or remain as p, I think the latter will be the case and so will need another look at in the near future.

Organisation main page now has strings externalised.

GAR IDs in the dashboard page were not read out very well by NVDA, so there are visually hidden tweaks to break up the GAR ID so NVDA at least reads it out in a more natural manner. This will need testing in other screen readers.

Messages for empty tables are now set as a html caption, a WAVE suggestion so screen readers know it is not a row.

The user registration used the cookie to populate the first and last name fields, which seemed odd. This is now changed so that on a validation error, the values are returned to the render context to be redisplayed. This is also reflected in the organisation invite user screen.

Added full stops after hints so the screen reader pauses in case of an error message, so it sounds more natural.

**Definition of Done**


nguyendhod:
This pull request will expect the updated E2E tests as well, as it will break current test build.

src/app/aircraft/index.njk
```diff
@@ -65,9 +65,7 @@
           </tr>
           {% endfor %}
         {% else %}
-          <tr class="govuk-table__row">
-            <td colspan="4"><p class="govuk-inset-text">{{__('aircraft_no_records')}}</p></td>
-          </tr>
+          {{ m.empty_table_message(__('aircraft_no_records')) }}
```
nguyendhod: All empty table messages now via a macro, and uses <caption>

src/app/garfile/arrival/index.njk
```diff
@@ -50,7 +50,7 @@
                   <input class="govuk-input govuk-date-input__input govuk-input--width-2" id="arrivalDay"
                     name="arrivalDay" type="number" pattern="[0-9]*" min="1" max="31"
                     value="{{cookie.dateSlice('day', cookie.getGarArrivalVoyage().arrivalDate)}}"
-                    aria-describedby="arrivalDate-error">
+                    {{ 'aria-describedby=arrivalDate-error' if errors | containsError('arrivalDate') }}>
```
nguyendhod: WAVE raised an error if 'arrivalDate-error' was set if there wasn't an actual span with id="arrivalDate-error" so this nunjuck filter goes through the error array and checks whether there is an element with identifier = 'arrivalDate' and sets the attribute if true.

This is pretty much everywhere where aria-describedby and errors are involved.

src/app/garfile/arrival/index.njk
```diff
@@ -132,12 +132,12 @@
           {{ m.error_message('arrivalPort', errors, __) }}
           <input class="govuk-input govuk-input--width-4" maxlength="4" id="arrivalPort" name="arrivalPort"
             autocomplete="off" type="text" value="{{cookie.getGarArrivalVoyage().arrivalPort | upper }}"
-            aria-describedby="arrivalPort-hint arrivalPort-error"/>
+            aria-describedby="arrivalPort-hint{{ ' arrivalPort-error' if errors | containsError('arrivalPort') }}"/>
         </div>
         {# End arrival port #}
 
         <div class="govuk-form-group">
-          <label class="govuk-heading-m">{{__('field_arrival_coordinates')}}</label>
```
nguyendhod: WAVE mentioned a label without a "for" is enough to be an error, so just changed to a span here.

src/app/garfile/departure/index.njk
```diff
@@ -22,64 +22,62 @@
...
-              <div class="govuk-date-input__item">
-                <div class="govuk-form-group">
-                  <label class="govuk-label govuk-date-input__label" for="departureMonth">
-                    <span class="govuk-visually-hidden">{{__('field_departure_hint')}}</span>{{__('field_month')}}
+            <div class="govuk-date-input__item">
+              <div class="govuk-form-group">
+                  <label class="govuk-label govuk-date-input__label" for="departureYear">
+                    <span class="govuk-visually-hidden">{{__('field_departure_hint')}}</span>{{__('field_year')}}
                   </label>
-                  <span id="departure-month-hint" class="govuk-visually-hidden">{{__('field_departure_month_hint')}}</span>
```
nguyendhod: Not sure what happened here, wrong indentation maybe?

src/app/home/index.njk
```diff
@@ -116,13 +116,13 @@
                 <td class="govuk-table__cell">{{ gar.departureTime }} </td>
                 <td class="govuk-table__cell">{{ gar.departurePort | upper }} </td>
                 <td class="govuk-table__cell">{{ gar.arrivalPort | upper }} </td>
-                <td id="garid-{{ name }}-{{ loop.index }}" class="govuk-table__cell">{{ gar.garId }} </td>
+                <td id="garid-{{ name }}-{{ loop.index }}" class="govuk-table__cell">
```
nguyendhod: Screen reader friendly output of the GAR ID. This is a macro as I sense this might be needed for other screens.

src/app/organisation/inviteusers/index.njk
```diff
@@ -30,77 +26,59 @@
...
       method="post" role="form">
 
-      <h1 class="govuk-heading-xl">Invite a user</h1>
-      <p class="govuk-body-l">Please begin by providing the following information to register a user account in your organisation.</p>
-
-       <div class="govuk-form-group {{ m.form_group_class('invite', errors) }}">
+      <h1 class="govuk-heading-xl">{{__('title_invite_user')}}</h1>
+      <p class="govuk-body-l">{{__('invite_user_caption')}}</p>
 
-          <div class="govuk-form-group {{ m.form_group_class('fname', errors) }}">
```
nguyendhod: Someone from a background of ASP.net perhaps?

Removed the spans regarding field-validation, and removed the width:100% as they aren't needed.

src/app/organisation/inviteusers/post.controller.js
```diff
@@ -18,6 +18,8 @@ module.exports = (req, res) => {
     .catch((err) => {
       logger.error('Invite Users Organisation postcontroller - There was a problem inviting a user');
       logger.error(JSON.stringify(err));
-      res.render('app/organisation/inviteusers/index', { cookie, errors: err });
```
nguyendhod: Return the supplied fname, lname and email so the form does not lose them on validation error

src/app/user/register/index.njk
```diff
@@ -7,14 +7,12 @@
 {% endblock %}
 
 {% block header %}
-{% include "header-pre-user.njk" %}
+  {% include "header-pre-user.njk" %}
```
nguyendhod: Not convinced there needs to be a pre-user and user header. Will need to refactor in due time.

src/app/user/register/index.njk
```diff
@@ -32,33 +30,33 @@
         <label class="govuk-label govuk-label--m" for="userFname">{{ __('field_given_name') }}</label>
         {{ m.error_message('userFname', errors, __) }}
         <input id="userFname" name="userFname" autocomplete="off" class="govuk-input govuk-input--width-30"
-          type="text" maxlength="{{ MAX_STRING_LENGTH }}" value="{{cookie.getUserFirstName()}}"
-          aria-describedby="userFname-error">
```
nguyendhod: Not sure why this ever used the cookie to populate in the first place. Now using variables passed by the controller on validation error (like the invite user screen)

src/app/user/register/post.controller.js
```diff
@@ -91,18 +91,18 @@ module.exports = (req, res) => {
...
-  cookie.setUserEmail(usrname);
-
   const validationChains = createValidationChains(fname, lname, usrname, cusrname);
 
   const isWhitelistRequired = (config.WHITELIST_REQUIRED.toLowerCase() === 'true');
 
   logger.info('Validating registration input');
   validator.validateChains(validationChains)
     .then(() => {
+      // Update the cookie
```
nguyendhod: Moved so that the cookie isn't upset unless validations pass.

src/app/user/register/post.controller.js
```diff
@@ -129,9 +129,11 @@ module.exports = (req, res) => {
     })
     .catch((err) => {
       logger.info('Failed registration validations');
-      logger.debug(JSON.stringify(err));
       res.render('app/user/register/index', {
         cookie,
+        fname,
```
nguyendhod: Passing said variables back to the form so that cookies aren't used to populate instead.

src/app/verify/mfa/index.njk
```diff
@@ -41,15 +41,16 @@
           </label>
           <span id="mfaCode-hint" class="govuk-hint">{{__('field_authentication_code_hint')}}</span>
           {{ m.error_message('mfaCode', errors, __) }}
-          <input id="mfaCode" autocomplete="off" class="govuk-input" maxlength="{{ mfaTokenLength }}" name="mfaCode" type="text" aria-describedby="mfaCode-hint mfaCode-error">
-          <p style="margin-top: 0.25em;" class="govuk-body-m">
-            <a href="/login/authenticate?resend=true">{{__('resend_code')}}</a>
-          </p>
+          <input id="mfaCode" autocomplete="off" class="govuk-input" maxlength="{{ mfaTokenLength }}" name="mfaCode" type="text" aria-describedby="mfaCode-hint{{ ' mfaCode-error' if errors}}">
```
nguyendhod: Moved the resend token button. I imagine a user with experience would tab and expect it to be the submit button only to find this one instead?

src/app/verify/mfa/post.controller.js
```diff
@@ -17,7 +21,7 @@ module.exports = (req, res) => {
   const cookie = new CookieModel(req);
 
   const mfaTokenLength = settings.MFA_TOKEN_LENGTH;
-  const errMsg = { message: 'There was a problem verifying your token. Try again' };
+  const errMsg = { identifier: 'mfaCode', message: i18n.__('validator_authentication_error') };
```
nguyendhod: Externalised the message, this will need applying to all validation messages!

src/common/templates/check-your-answers.njk
```diff
@@ -2,149 +2,66 @@
...
-        <dd class="app-check-your-answers__change"></dd>
-      </div>
-      <div class="app-check-your-answers__contents">
-        <dt class="app-check-your-answers__question">Departure port coordinates</dt>
-        <dd class="app-check-your-answers__answer">
-          {{__('field_latitude')}}: {{garfile.departureLat}}<br>
-          {{__('field_longitude')}}: {{garfile.departureLong}}</dd>
-        <dd class="app-check-your-answers__change">
-        </dd>
-      </div>
```
nguyendhod: Using a macro because the rows are all virtually the same, this reduces a lot of the markup on this page.

src/common/templates/includes/header-user.njk
```diff
@@ -15,8 +15,7 @@
       </a>
     </div>
     <div class="govuk-header__content">
-
-      <a href="/home" class="govuk-header__link govuk-header__link--service-name">{{__('title')}}</a>
+      <a href="/" class="govuk-header__link govuk-header__link--service-name">{{__('title')}}</a>
```
nguyendhod: Original story mentioned that clicking this was essentially a log out because you then had to go to the login page when you clicked "Start Now" on the home page.

This is now not the case if a session is still present, so this is a bug that doesn't apply anymore.

WAVE kicks up a fuss that there are two adjacent links going to the same place, hence the want to change this.

src/common/templates/includes/macros.njk
```diff
@@ -16,6 +16,12 @@
 	{% endfor %}
 {% endmacro %}
 
+{% macro empty_table_message(message) %}
```
nguyendhod: WAVE mentions a piece of text that part of a table but not a cell should be marked as a caption or somesuch.

This is now the case, with styling to set it to the bottom, which looks ok to be honest. Not sure what effect this might have on newer browsers (was not aware of a caption tag before).

src/common/utils/templateFilters.js
```diff
@@ -7,4 +7,12 @@ const uncamelCase = (value) => {
   return uncamelStr;
 };
 
+const containsError = (array, value) => {
```
nguyendhod: This is the logic for the aria-describedby for errors.

Goes through the error array, usually elements are an identifier/message pair, so check the identifier for a match, and return true if so. This will then add to the attribute back in the original template.

src/locales/en.json
```diff
@@ -33,7 +33,7 @@
     "nav_back": "Back",
 
     "field_username": "Registered email address",
-    "field_username_hint": "Enter the email address you used when you created your account",
```
nguyendhod: Added full stops to all hints, mainly because if an error message is also present, the screen reader reads it all as one sentence, where a small break to inform user of an error is desirable.

src/test/common/services/sendEmail.test.js
```diff
@@ -26,9 +26,12 @@ describe('Send Email Service', () => {
     sinon.restore();
   });
 
+  // For the sake of code coverage, have this rewire the logger for production...
   it('should throw an error if no API key set', () => {
+    process.env.NODE_ENV = 'production';
```
nguyendhod: This does not fully test what happens when the logger is in 'production'. Purely for code coverage purposes.

src/test/user/register/post.controller.test.js
```diff
@@ -61,18 +63,35 @@ describe('User Register Post Controller', () => {
...
       },
     };
-    try {
+    const cookie = new CookieModel(emptyRequest);
+
+    callController = async () => {
       await controller(emptyRequest, res);
-    } catch (err) {
-      expect(err).to.eq('Validation error when logging in');
-      expect(res.render).to.have.been.calledWith('app/user/login/index');
```
nguyendhod: Made the assertion here much stronger, hence the sudden addition of exact parameters.

PR #117, opened 2019-08-08
Add screen reader tweaks to aircraft pages

nguyendhod:
Cleaned markup, added ids to hints and errors and the aria-describedby attributes.

Added the revised submission buttons.

Added javascript tweak to add buttonClicked field for the "Add to GAR" button, which may need applying for the manifest page.

Use of the updated form submissions will break the current E2E tests.

Initial pass of the aircraft screens, both saved and for GAR forms.

**Definition of Done**


nguyendhod:
Not sure if the E2E tests will pass, given the issues with the organisation gmail account...

PR #121, opened 2019-08-13
SGAR-797 unregistered users

nguyendhod:

When an unregistered user attempts to log into the system, they are presented with an error message and a link to the register account screen instead of going straight to the token verification.

The Cookies page is also externalised into the i18n library. There is also some SonarQube driven development where duplicate blocks of code were addressed. Enough unit and E2E tests to hopefully prevent regressions when doing this.

**Definition of Done**


src/app/cookie/index.njk
```diff
@@ -1,159 +1,131 @@
...
+      <li>{{__('how_cookies_are_used_1_list_item_1')}}</li>
+      <li>{{__('how_cookies_are_used_1_list_item_2')}}</li>
+      <li>{{__('how_cookies_are_used_1_list_item_3')}}</li>
+      <li>{{__('how_cookies_are_used_1_list_item_4')}}</li>
+    </ul>
 
-  <p class="govuk-body-m">
-    In order to improve usability and retain information about your account and privileges, we store a unique identifier for each session
-  </p>
+    {% macro table_header() %}
```
nguyendhod: Created a pair of macros to create the header (which is the same markup for all the tables) and the rows (taking in the 3 column values as parameters)

src/app/cookie/index.njk
```diff
@@ -1,159 +1,131 @@
...
-		Cookies
-		</h1>
+		<h1 class="govuk-heading-xl">{{__('title_cookies')}}</h1>
 
 		<p class="govuk-body-m">
-      This service puts small files (known as 'cookies') onto your computer to collect
-      information about how you browse the site.
-      <br><br>
-      Cookies are used to:
+      {{__('cookies_caption')}}
```
nguyendhod: Lots of externalising the strings into the i18n package.

PR #122, opened 2019-08-15
SGAR-769 navigation links highlighted

nguyendhod:

Uses variables set in nunjucks template to determine which navigation link should be highlighted, with the GOV.UK design approach for highlighting the link used. No unit tests as template code change mainly (could explore this), tests are within the E2E tests.

**Definition of Done**


src/app/aircraft/add/index.njk
```diff
@@ -6,8 +6,10 @@
 	{{__('error_prefix') if errors}} GAR | Save an Aircraft
 {% endblock %}
 
+{% set active = '/aircraft' %}
```
nguyendhod: This nunjucks "active" variable is used by the header links as the logic to determine whether it should be highlighted (set to active), by comparing it to the href it points to.

Simplistic, but reasonable efficient approach.

src/app/organisation/editorganisation/post.controller.js
```diff
@@ -19,13 +19,15 @@ module.exports = (req, res) => {
...
       // API should return OrgId
-      cookie.setOrganisationName(req.body.orgName);
       organisationApi.update(orgName, cookie.getOrganisationId())
         .then((apiResponse) => {
           const responseObj = JSON.parse(apiResponse);
           logger.debug(`Response from API: ${JSON.stringify(responseObj)}`);
           // Check for error
-          res.redirect('/organisation');
+          cookie.setOrganisationName(req.body.orgName);
+          req.session.save(() => {
```
nguyendhod: Sometimes, the E2E tests would perform edit organisation save operations quickly enough that the session is not updated before the test assertions.

Added the req.session.save with the redirect as the callback to ensure the page returns only after the session is saved (and the cookie value is set).

src/common/services/tokenApi.js
```diff
@@ -29,15 +29,12 @@ module.exports = {
       }, (error, response, body) => {
         if (error) {
           logger.error('There was a problem calling the settoken API');
-          return console.dir(error);
+          logger.error(error);
+          reject(error);
```
nguyendhod: Potentially, if the endpoint ever rejected for any reason, then this Promise would be unhandled. There is no particular reason why this would not mean the application could have hung. Error is now wrapped into a reject call.

Other service calls may need looking at to ensure this is not the case.

src/common/services/tokenApi.js
```diff
@@ -29,15 +29,12 @@ module.exports = {
...
         }
         logger.debug('Successfully called settoken API');
         resolve(body);
-        return body;
       });
-    }).catch((err) => {
```
nguyendhod: This catch block does not get triggered, as the code just returns a Promise, not invoke it.

src/common/templates/includes/header-user.njk
```diff
@@ -18,33 +18,24 @@
...
 
       <button type="button" role="button" class="govuk-header__menu-button js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
+      {% macro nav_link(id, href, label) %}
+        <li id="nav-{{ id }}-li" class="govuk-header__navigation-item{{ ' govuk-header__navigation-item--active' if active == href}}">
+          <a id="nav-{{ id }}" class="govuk-header__link" href="{{ href }} ">{{ label }}</a>
+        </li>
+      {% endmacro %}
       <nav>
         <ul id="navigation" class="govuk-header__navigation" aria-label="Top Level Navigation">
-          <li class="govuk-header__navigation-item">
```
nguyendhod: The links are now created with the above macro to reduce the code. Macro also contains the logic for the active determination.

src/locales/en.json
```diff
@@ -329,7 +330,15 @@
     "invite_user_given_name_hint": "Enter the given name of the user.",
     "invite_user_surname_hint": "Enter the surname of the user.",
     "invite_user_email_hint": "Enter the email of the user.",
-	"invite_user_cemail_hint": "Confirm the email of the user.",
+    "invite_user_cemail_hint": "Confirm the email of the user.",
```
nguyendhod: Invite user externalised strings.

src/test/common/services/tokenApi.test.js
```diff
@@ -51,6 +54,19 @@ describe('UserSessions', () => {
       });
   });
 
+  it('setMfaToken rejects', () => {
```
nguyendhod: These particular reject tests might need applying to a lot of the service tests, to increase code coverage.

src/app/garfile/manifest/addnewperson/index.njk
```diff
@@ -16,9 +18,6 @@
 {% endblock %}
 
 {% block content %}
-
-  {% import "common/templates/includes/macros.njk" as m %}
```
abrarhaidari: What was this import used for before?

abrarhaidari: Duplicated import - ok to remove

PR #123, opened 2019-08-15
SGAR-1029 GAR Summary user check

nguyendhod:

A check is now in place when entering the GAR Summary Page, which checks the requested GAR and determines whether the currently logged in user should be allowed to see the GAR. They are redirected back to the dashboard if they are not.

This is a breaking change, which will require the equivalent updates in the data-access-api to be merged for this to work.

**Definition of Done**


nguyendhod:
The description probably did not emphasise this:

WARNING: This is a breaking change, the data-access-api updates must be in place otherwise app functionality will be affected!

PR #124, opened 2019-08-21
SGAR-767 port radio button

nguyendhod:
Departure and arrival ports screen now use a radio button to determine whether the port code or coordinates are used.

Behind the scenes, the port is set to 'ZZZZ' as appropriate, with the summary and review screens also hiding and showing the port or coordinates as appropriate.

This requires the E2E tests to be merged prior as it is a breaking change.

**Definition of Done**


src/app/garfile/arrival/index.njk
```diff
@@ -57,7 +57,7 @@
                   <label class="govuk-label govuk-date-input__label" for="arrivalMonth">
                     <span class="govuk-visually-hidden">{{__('field_arrival_hint')}}</span>{{__('field_month')}}
                   </label>
-                  <span id="arrival-month-hint" class="govuk-visually-hidden">{{__('field_departure_month_hint')}}</span>
+                  <span id="arrival-month-hint" class="govuk-visually-hidden">{{__('field_month_hint')}}</span>
```
nguyendhod: This was actually missing! So screen reader was probably reading "field departure month hint" out. Whoops

src/app/garfile/arrival/index.njk
```diff
@@ -184,5 +194,6 @@
       selectElement: document.querySelector('#arrivalPort'),
       minLength: 2,
     });
+    document.getElementById('arrivalPort').setAttribute('aria-describedby', 'arrivalPort-hint{{ ' arrival-error' if errors | containsError('arrivalPort') }}');
```
nguyendhod: Javascript here to add the aria-describedby attribute to the accessible autocomplete. This is something that probably the component itself should have done anyway...

src/app/garfile/departure/index.njk
```diff
@@ -186,6 +207,7 @@
     selectElement: document.querySelector('#departurePort'),
     minLength: 2,
   });
+  document.getElementById("departurePort").setAttribute('aria-describedby', 'departurePort-hint{{ ' departurePort-error' if errors | containsError('departurePort') }}');
```
nguyendhod: Javascript here to add the aria-describedby attribute to the accessible autocomplete. This is something that probably the component itself should have done anyway...

src/app/garfile/arrival/index.njk
```diff
@@ -120,56 +120,66 @@
         </div>
         {# End arrival time #}
 
-        {# Start arrival port #}
```
nguyendhod: Essentially took the port autocomplete and coordinate divs and placed them as containers within the radio group conditional, akin to how it works in the customs page.

This and the departure nunjucks is the same, so possibly could have templated them, if not for that being far too complicated than it was worth.

src/app/garfile/arrival/post.controller.js
```diff
@@ -116,8 +107,15 @@ module.exports = async (req, res) => {
 
   // Define voyage
   const voyage = req.body;
-  voyage.arrivalPort = _.toUpper(voyage.arrivalPort);
   delete voyage.buttonClicked;
+  if (voyage.portChoice === 'No') {
```
nguyendhod: Looks at the choice made and sets values as appropriate.
To clear the coordinates, empty strings are used, as it appears undefined or null get ignored when sent to the back end.

src/app/garfile/departure/post.controller.js
```diff
@@ -7,7 +7,7 @@ const ValidationRule = require('../../../common/models/ValidationRule.class');
 
 const createValidationChains = (voyage) => {
   // Define port / date validation msgs
-  const portMsg = 'As you have entered a departure port code of "YYYY", you must provide longitude and latitude coordinates for the location';
```
nguyendhod: Redundant validation now, replaced with the validation for ensuring a choice is made.

src/app/garfile/departure/post.controller.js
```diff
@@ -49,16 +46,16 @@ const createValidationChains = (voyage) => {
...
 
   // Check if port code is ZZZZ then need to validate lat/long and display req zzzz message
   if (departPortObj.portCode.toUpperCase() === 'YYYY') {
     validations.push(
-      departurePortZZZZ,
       departureLatValidation,
       departureLongValidation,
     );
-  } else {
+  } else if (voyage.portChoice) {
```
nguyendhod: If this was not changed, then the validations would return both the port choice not being made and the departure port not being set, when only the first one makes sense to a user.

src/app/home/index.njk
```diff
@@ -34,9 +34,6 @@
     {% endif %}
 	</div>
 </div>
-
```
nguyendhod: Duplicate macro import.

src/app/home/index.njk
```diff
@@ -116,8 +113,8 @@
                 </td>
                 <td class="govuk-table__cell">{{ gar.departureDate }}</td>
                 <td class="govuk-table__cell">{{ gar.departureTime }}</td>
-                <td class="govuk-table__cell">{{ gar.departurePort | upper }}</td>
-                <td class="govuk-table__cell">{{ gar.arrivalPort | upper }}</td>
+                <td class="govuk-table__cell">{{ __('value_port_code_zzzz') if gar.departurePort == 'YYYY' else gar.departurePort | upper }}</td>
```
nguyendhod: Set the label to be "N/A" if not port. More for consistency as 'YYYY' nor 'ZZZZ' are referenced in the application itself anymore.

src/common/services/garApi.js
```diff
@@ -3,105 +3,104 @@ const logger = require('../utils/logger')(__filename);
...
-          return 'Failed to update GAR';
+          reject(error);
+          return;
         }
-        resolve(body);
         logger.debug('Successfully called GAR put endpoint');
-        return body;
-      });
-    })
-      .catch((err) => {
```
nguyendhod: Removed these catch blocks as they actually could hang. Same with adding in rejects. This is much more in line with Promises should have been done in the first place.

The hope is that more useful error messages will be logged.

src/common/templates/check-your-answers.njk
```diff
@@ -13,12 +13,44 @@
         {% endif %}
       </div>
     </dl>
+
```
nguyendhod: Moved the macros from the macros.njk to here because it is only used for this screen.

src/common/templates/check-your-answers.njk
```diff
@@ -13,12 +13,44 @@
...
+  
+    {% macro check_your_answers_row(label, value) %}
+      <div class="app-check-your-answers__contents">
+        <dt class="app-check-your-answers__question">{{ label }}</dt>
+        <dd class="app-check-your-answers__answer">{{ value }}</dd>
+      </div>
+    {% endmacro %}
+
+    {% macro check_your_answers_row_port_code(id, label, value) %}
+      {% if value =='ZZZZ' or value =='YYYY' %}
```
nguyendhod: Eventually, the YYYY could be removed, but doesn't matter. Renders nothing if it is the value to represent coordinates being set.

src/common/templates/check-your-answers.njk
```diff
@@ -13,12 +13,44 @@
...
+      {% else %}
+        <div class="app-check-your-answers__contents">
+          <dt id="{{ id }}-label" class="app-check-your-answers__question">{{ label }}</dt>
+          <dd id="{{ id }}-value" class="app-check-your-answers__answer">{{ value }}</dd>
+        </div>
+      {% endif %}
+      
+    {% endmacro %}
+
+    {% macro check_your_answers_row_coordinates(id, label, latitude, longitude, __) %}
```
nguyendhod: Added id parameter to make E2E scripts easier, had to use xpath otherwise.

This should probably be applied to the normal row macro as well.

src/locales/en.json
```diff
@@ -130,8 +131,13 @@
     "field_arrival_port": "Arrival port",
     "field_arrival_coordinates": "Arrival coordinates",
 
-    "field_port_hint": "Provide the IATA or ICAO code for the port. If the port does not have a code, enter '%s'. The arrival port must be different to the departure port.",
-    "field_coordinate_hint": "You are only required to enter coordinates if your port does not have an IATA or ICAO code. Provide coordinates in decimal format to 4 decimal places.",
+    "field_port_choice": "Do you know the departure port's ICAO or IATA code?",
+    "field_choice_yes": "Yes",
+    "field_choice_no": "No",
+    "field_port_code": "IATA or ICAO code",
+    "field_port_choice_hint": "If the International Civil Aviation Organisation (ICAO) or International Air Transport Association (IATA) code is unknown, geographic coordinates are expected.",
```
nguyendhod: An example of the explain abbreviation on first use element of the site. Whether this is to be adopted, or whether it will become a separate screen of terms, business still need to discuss.

src/common/middleware/usercheck.js
```diff
@@ -1,16 +1,24 @@
 const CookieModel = require('../../common/models/Cookie.class');
 const logger = require('../utils/logger')(__filename);
 
+const checkUserInCookie = (cookie) => {
```
nguyendhod: User check middleware now factors in the verified status of the user and the role of the user in the cookie. If they are not there, then the app considers the user not logged in.

src/app/user/login/get.controller.js
```diff
@@ -1,7 +1,7 @@
 const navUtil = require('../../../common/utils/nav');
 
 module.exports = (req, res) => {
-  if (req.headers.referer && req.session.u && req.session.u.dbId) {
+  if (req.headers.referer && req.session.u && req.session.u.dbId && req.session.u.vr && req.session.u.rl) {
```
nguyendhod: Akin to the usercheck, also check whether verified and role values are set.

abrarhaidari:
Tested locally, unit test and e2e tests pass.

PR #127, opened 2019-08-22
SGAR-1069 GAR aircraft pagination CSRF issue

nguyendhod:
Clash of the disableAutoFill library, where it scrambles form names, in this case, scrambling the csrf form name, which is then not picked up by the csrf checking middleware, as links are used which do not execute the unscrambling.

Simplest option is to not use the library, which should be acceptable for this particular screen.

Most effort has been elsewhere to further improve unit test code coverage.

**Definition of Done**


src/app/api/uploadfile/post.controller.js
```diff
@@ -31,26 +31,33 @@ const exceedFileLimit = (fileSize, garId) => {
...
+      logger.error(deleteSupportingDocErr);
+      res.redirect('/garfile/supportingdocuments?query=deletefailed');
+    });
+  return true;
+};
+
 module.exports = (req, res) => {
   logger.info('Entering upload file post controller');
 
-  if (req.body.deleteDocId) {
```
nguyendhod: Moved the delete block into its own method to reduce the main method size.

src/app/garfile/cancel/index.js
```diff
@@ -1,29 +1,7 @@
-// Npm dependencies
-const express = require('express');
-
-// Middleware
-const flagpole = require('../../../common/middleware/flagpole');
-const usercheck = require('../../../common/middleware/usercheck');
-const csrfcheck = require('../../../common/middleware/csrfcheck')
-const parseForm = require('../../../common/middleware/parseForm');
+const nav = require('../../../common/utils/nav');
```
nguyendhod: Using the refactored common router creation

src/app/garfile/craft/index.njk
```diff
@@ -25,7 +25,7 @@
 		{% include "common/templates/includes/errors.njk" %}
 	{% endif %}
 
-  <form id="page-form" action="/garfile/craft" autocomplete="off" class="form-group" method="post" role="form">
+  <form id="previous_aircraft" action="/garfile/craft" autocomplete="off" class="form-group" method="post" role="form">
```
nguyendhod: The "page-form" id is picked up by the disableAutoFill library, so changing the id is essentially the way to "disable" it.

src/app/garfile/craft/index.njk
```diff
@@ -87,7 +87,7 @@
         <label class="govuk-label" for="craftReg">{{__('field_aircraft_registration')}}</label>
         {{ m.hint('craftReg', __('field_aircraft_registration_hint'))}}
         {{ m.error_message('craftReg', errors, __) }}
-        <input id="craftReg" name="craftReg" class="govuk-input govuk-input--width-10"
+        <input id="craftReg" name="craftReg" class="govuk-input govuk-input--width-10" autocomplete="off"
```
nguyendhod: Added these autocomplete="off" to try and prevent some sort of autocomplete capability.

src/app/register/logout/get.controller.js
```diff
@@ -4,6 +4,6 @@ module.exports = (req, res) => {
   const cookie = new CookieModel(req);
   req.session.destroy(() => {
     cookie.reset();
-    res.redirect('/login');
+    res.redirect('/welcome/index');
```
nguyendhod: When logging out, just goes to the home page instead of the login page. Matter of preference really?

src/app/user/login/get.controller.js
```diff
@@ -1,7 +1,20 @@
...
+const checkUserInSession = (userSessionObject) => {
+  if (userSessionObject) {
+    return userSessionObject.dbId && userSessionObject.vr && userSessionObject.rl;
+  }
+  return false;
+};
+
 module.exports = (req, res) => {
-  if (req.headers.referer && req.session.u && req.session.u.dbId && req.session.u.vr && req.session.u.rl) {
+  if (req.headers.referer && checkUserInSession(req.session.u)) {
```
nguyendhod: Sonarqube driven development, moving the boolean predicates into their own function.

src/app/user/login/post.controller.js
```diff
@@ -76,7 +76,7 @@ module.exports = (req, res) => {
       });
     })
     .catch((err) => {
-      logger.error('Validation error when logging in');
+      logger.info('Validation error when logging in');
```
nguyendhod: Validation errors are not errors in terms of exceptions etc, so normally they are logged at INFO level.

src/common/services/craftApi.js
```diff
@@ -23,21 +23,16 @@ module.exports = {
           craftType,
           craftBase,
         }),
-      }, (error, response, body) => {
+      }, (error, _response, body) => {
```
nguyendhod: Linter complains of an unused parameter, so the underscore is added to bypass that.

src/common/services/craftApi.js
```diff
@@ -23,21 +23,16 @@ module.exports = {
...
         if (error) {
           logger.error('Failed to call craft creation API');
           reject(error);
-          return error;
```
nguyendhod: Promises seemed to have previously returned the error or the body, which is not really expected. Reject or resolve calls are, so these have changed.

src/common/templates/includes/pagination.njk
```diff
@@ -22,7 +22,7 @@ passed in explicitly it seems.
     <ul class="pagination">
     {% if pages.totalPages > 1 %}
       {% if pages.currentPage > 1 %}
-        <li class="pagination__item"><a id="{{formName}}_pagination_previous" class="pagination__link" href="#" onclick="transitionToPage({{ pages.currentPage - 1 }})" aria-label="Previous page"><span aria-hidden="true" role="presentation">&laquo;</span>{{__('pagination_previous')}}</a></li>
```
nguyendhod: Removed these hrefs as they are not needed.

src/common/utils/nav.js
```diff
@@ -20,7 +20,9 @@ const buildRouterAndPaths = (path, getController, postController) => {
 
   // Routing
   router.get(paths.index, flagpole, usercheck, csrfcheck, getController);
-  router.post(paths.index, flagpole, usercheck, parseForm, csrfcheck, postController);
+  if (postController) {
```
nguyendhod: Some of the index.js router definitions do not define a POST controller, so this is now optional and so means no overloaded functions are needed.

src/test/common/services/craftApi.test.js
```diff
@@ -6,11 +6,15 @@ const chai = require('chai');
 const sinonChai = require('sinon-chai');
 const proxyquire = require('proxyquire');
 const { URL } = require('url');
+const nock = require('nock');
```
nguyendhod: crafts.test.js used to be in the test root folder, which used nock to mock HTTP calls. This is just as viable a means to test as the use of proxyquire and sinon mocks here, so they have been combined into a single relevant class.

The nock approach also makes testing errors easier and thus checks for the Promise reject path can be done.

src/test/common/services/createUserApi.test.js
```diff
@@ -5,16 +5,92 @@ const { expect } = require('chai');
 const chai = require('chai');
 const sinonChai = require('sinon-chai');
 const proxyquire = require('proxyquire');
-const { URL } = require('url');
+const nock = require('nock');
 
 require('../../global.test');
+const endpoints = require('../../../common/config/endpoints');
+const registerApi = require('../../../common/services/createUserApi');
```
nguyendhod: This originally was in users.test.js, and whilst true in a sense, it makes more sense to group tests by the class they are in, so the createUserApi based nock tests are now here.

src/test/user/login/get.controller.test.js
```diff
@@ -43,6 +43,7 @@ describe('User Login Get Controller', () => {
   it('should render if referer but no user object in session', async () => {
     req.headers.referer = '/example';
     const cookie = new CookieModel(req);
+    delete req.session.u;
```
nguyendhod: Controller code never had the situation of no user defined tested.

It is actually a contrived scenario, as a user object is always present due to the CookieModel creating it, even with null values when empty.

nguyendhod:
Merged with master, so ready for review.

PR #133, opened 2019-10-16
Changed the version of clamav to 'latest' so that it can pick up the …

srinisundar:
…most up-to-date image version.

**Definition of Done**


berhanep:
It's fine to go ahead and update the clamav version to the latest.

PR #294, opened 2023-04-03
date validation errrors suggest correct cell format

ho-henry-senior:
added text to file upload date validation message to suggest to user they check date fields are formatted as dates


sjose012 (approved):
Looks good

PR #300, opened 2023-04-12
Feature/bed 13

ho-henry-senior:

**Definition of Done**

### DIFF
--- a/src/app/garfile/arrival/post.controller.js
+++ b/src/app/garfile/arrival/post.controller.js
@@ -86,7 +86,6 @@ const buildValidations = (voyage) => {
   const validations = [
     [new ValidationRule(validator.realDate, 'arrivalDate', arriveDateObj, realDateMsg)],
     [new ValidationRule(validator.currentOrFutureDate, 'arrivalDate', arriveDateObj, futureDateMsg)],
-    // [new ValidationRule(validator.dateNotTooFarInFuture, 'arrivalDate', arriveDateObj, __('field_arrival_date_too_far_in_future'))],
     [new ValidationRule(validator.validTime, 'arrivalTime', arrivalTimeObj, timeMsg)],
     [new ValidationRule(validator.notEmpty, 'portChoice', voyage.portChoice, portChoiceMsg)],
   ];
--- a/src/app/garfile/arrival/post.controller.js
+++ b/src/app/garfile/arrival/post.controller.js
@@ -86,7 +86,7 @@ const buildValidations = (voyage) => {
   const validations = [
     [new ValidationRule(validator.realDate, 'arrivalDate', arriveDateObj, realDateMsg)],
     [new ValidationRule(validator.currentOrFutureDate, 'arrivalDate', arriveDateObj, futureDateMsg)],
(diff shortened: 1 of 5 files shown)


src/common/utils/validator.js
```diff
@@ -217,18 +217,31 @@ function currentOrFutureDate(dObj) {
   return false;
 }
 
-// function dateTooFarInFuture(dObj){
-//   if (dObj === null || dObj === undefined) return false;
+function dateNotTooFarInFuture(dObj) {
```
anushasivakumar, line 220: Please can we add a docstring for what this function does

src/common/utils/validator.js
```diff
@@ -217,18 +217,31 @@ function currentOrFutureDate(dObj) {
...
+
+  if (dObj instanceof Date) {
+    providedDate = dObj;
+  } else {
+
+    if (!(numericDateElements(dObj)
+      && validDay(dObj.d, dObj.m, dObj.y)
+      && validMonth(dObj.m)
+      && validYear(dObj.y))) {
+      return false;
```
anushasivakumar, line 237: Can we verify this check, please - currently this is returning false if it is not a numericvalue but a valid date/month/day

src/test/validators.test.js
```diff
@@ -816,4 +817,44 @@ describe('Validator', () => {
...
+      expect(actualResult).to.equal(false);
+    });
+
+    it('Should accept a date within one month in the future in object format', () => {
+      actualResult = validator.dateNotTooFarInFuture({ d: '11', m: '5', y: '2023' });
+      expect(actualResult).to.equal(true);
+    });
+
+    it('Should reject a date further than one month in the future in date format', () => {
+      actualResult = validator.dateNotTooFarInFuture(new Date(2024, 0, 1));
```
anushasivakumar, line 847: Is this test for an invalid date or for one month in the future in date format? The month seems to be '0'. Perhaps we need tests for both the cases.

anushasivakumar (commented):
Overall looks good, a few minor comments added.

src/app/garfile/arrival/post.controller.js
```diff
@@ -86,7 +86,7 @@ const buildValidations = (voyage) => {
   const validations = [
     [new ValidationRule(validator.realDate, 'arrivalDate', arriveDateObj, realDateMsg)],
     [new ValidationRule(validator.currentOrFutureDate, 'arrivalDate', arriveDateObj, futureDateMsg)],
-    // [new ValidationRule(validator.dateTooFarInFuture, 'arrivalDate', arriveDateObj, __('field_arrival_date_too_far_in_future'))],
+    // [new ValidationRule(validator.dateNotTooFarInFuture, 'arrivalDate', arriveDateObj, __('field_arrival_date_too_far_in_future'))],
```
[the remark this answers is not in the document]
ho-henry-senior: ok will remove

src/common/utils/validator.js
ho-henry-senior: will do

src/common/utils/validator.js
anushasivakumar: It is good to have the check but we just need to double check if the condition is correct.

ho-henry-senior: have confirmed with Anusha that condition is correct

anushasivakumar (approved):

PR #301, opened 2023-04-13
added anchore scan to pipeline

JHoldergov:

Integrating Anchore scanning into the pipeline.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,24 +1,37 @@
 ---
 kind: pipeline
-name: default
+name: image-scan-vulnerabilities
 type: kubernetes
 
-services:
-- name: docker
-  image: 340268328991.dkr.ecr.eu-west-2.amazonaws.com/acp/dind:latest
+
+platform:
+  os: linux
+  arch: amd64
+
 
 steps:
-- name: Build
+- name: build_and_push
   pull: if-not-exists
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -44,29 +57,54 @@ steps:
...
-  - docker push quay.io/ukhomeofficedigital/egar-public-site-ui:$${DRONE_COMMIT}
+    IMAGE_NAME: quay.io/ukhomeofficedigital/egar-public-site-ui:${DRONE_COMMIT}
+    TOLERATE: high
   when:
-    branch:
-    - master
-    - hotfix/*
     event:
+    - pull_request
     - push
```
anushasivakumar, line 70: Do we need to run the scan for every commit ? Perhaps after a PR is created is sufficient? Happy to know the reason behind this.

.drone.yml
```diff
@@ -44,29 +57,54 @@ steps:
 #    - push
 #    - pull_request
 
-- name: Push Image to Quay.io
-  pull: if-not-exists
-  image: 340268328991.dkr.ecr.eu-west-2.amazonaws.com/acp/dind:latest
+- name: maintenance-image-scan
```
anushasivakumar, line 60: Can we please rename this to something more meaningful that explains the purpose of the scan.

anushasivakumar (changes requested):
Thank you for the changes, have added a couple of comments.

PR #317, opened 2023-04-20
BED-28 Fix 'Not able to delete a user from eGAR'

sjose012:

**Definition of Done**


src/common/config/endpoints.js
```diff
@@ -26,7 +26,7 @@ const endpoints = {
     return endpoint;
   },
   deleteUser(userId, email) {
-    const endpoint = new URL(`${API_VERSION}/user/${userId || email}`, BASE_URL).href;
+    const endpoint = new URL(`${API_VERSION}/user/${userId, email}`, BASE_URL).href;
```
ho-henry-senior, line 29: I am sorry about this, I seems I have introduced this bug when attempting to fix unit tests.
The code, as implemented, seemed wrong to me. This is a description of the js comma operator from [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comma_operator):

```The comma (,) operator evaluates each of its operands (from left to right) and returns the value of the last operand.```

This being the case, only `email` would ever be returned; the operands are all primitives so evaluating them would have no side-effects.

The above code broke a unit test that my change fixed, and my change seemed like a more sensible implementation - append either `userId` or `email` as param to `/user/` endpoint depending on which was supplied.

I don't understand how my code broke things and I still don't understand what the above implementation is mean to do, but am sorry for having broken the code. I will have a better look at what was wrong tomorrow.

PR #327, opened 2023-05-09
Feature/bed 27 fix unit tests take 4

ho-henry-senior:
- Fixes all unit tests except for the supertest ones. I can't figure out how to get them to work.
- Refactors user deletion and organisation deletion - it basically undoes some of Paulos' work.
It seemed to work fine to me but I might need to check with @sjose012 who has also worked on this.
- Beyond this the only change to application code (vs test code) was to `src/app/unavailable/get.controller.js` - to get the unit tests to work I had to put conditional logic inside a single `exports` statement rather then conditionally rendering one of two `export` statements.

**Definition of Done**


ame-test:
Henry mentioned Sunish was need to review due changes he worked on.

Went through the code with Henry:
* Worked in local.
* Known not to work in docker compose due to `npm install --production` flag.
* drone unit test section disabled is known.

sjose012 (approved):

PR #330, opened 2023-06-21
Bed 55 import codes

ho-henry-senior:
Updated airport file 
Fixed script that generates airport file (although the updates to the airport file did not use the script - the datasource the script uses did not have the new airports).

**Definition of Done**

### DIFF
--- a/src/common/utils/airport_codes.json
+++ b/src/common/utils/airport_codes.json
@@ -1844,6 +1844,12 @@
     "label": "Liège Airport (Belgium) (EBLG / LGG)"
   },
   {
+    "id": "EBMB",
+    "id2": "EBMB",
+    "british": false,
+    "label": "Melsbroek Air Base (Belgium) (EBMB)"
+  },
+  {
     "id": "OST",
     "id2": "EBOS",
     "british": false,
--- a/src/common/utils/airport_codes.json
+++ b/src/common/utils/airport_codes.json
@@ -46253,7 +46253,13 @@
     "id": "EGCA",
     "id2": "EGCA",
     "british": true,
-    "label": "COAL ASTON - BENTLEY FARM (United Kingdom) (EGCA)"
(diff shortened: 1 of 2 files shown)


src/common/utils/airport_codes.json
```diff
@@ -46253,7 +46253,13 @@
...
     "id2": "EGCA",
     "british": true,
-    "label": "COAL ASTON - BENTLEY FARM (United Kingdom) (EGCA)"
+    "label": "Coal Aston Airfield / Apperknowle Airstrip (United Kingdom) (EGCA)"
+  },
+  {
+    "id": "COAL ASTON - BENTLEY FARM",
+    "id2": "COAL ASTON - BENTLEY FARM",
+    "british": true,
+    "label": "COAL ASTON - BENTLEY FARM (United Kingdom)"
```
anushasivakumar, line 46262: Not sure about the significance of using `/`, Do we need that here as well?

ho-henry-senior: Sonali did say she thought the `/` played a syntactic role, but, looking at the rest of the file, it doesn't seem to be used consistently  (I left Bentley Farm formatted as I found it)

for e.g.

```json 
 {
    "id": "EGBR",
    "id2": "EGBR",
    "british": true,
    "label": "BREIGHTON - SELBY (United Kingdom) (EGBR)"
  },
```

I could change it to `COAL ASTON / BENTLEY FARM (United Kingdom)` but it's not clear to me if it's right/needed, and it begs the question of whether to do this throughout the file? I am trying to get an explanation from CBP.

I think where the `/` might be significant is in the brackets with the code(s):
```json 
{
    "id": "KJK",
    "id2": "EBKT",
    "british": false,
    "label": "Wevelgem Airport (Belgium) (EBKT / KJK)"
 },
```

hoping CBP can clear this up for me.

sjose012 (approved):

PR #347, opened 2023-08-25
Country code for response_country instead of free text

ame-test:
- AMG requires "operatorCountry" to be a country code (see ICD document 1.1 page 30).
- Changes so the responsibleCountry value is a country code rather than free text.

## Main files
- `./garfile/responsibleperson/index.njk`

**Definition of Done**


src/test/garfile/responsibleperson/get.controller.test.js
```diff
@@ -59,7 +59,7 @@ describe('GAR Responsible Person Get Controller', () => {
       responsibleAddressLine2: 'United Federation of Planets',
       responsibleTown: 'Alpha Quadrant',
       responsiblePostcode: 'NCC-1701D',
-      responsibleCounty: 'Earth',
+      responsibleCounty: 'GBR',
```
ho-henry-senior, line 62: should this be `responsibleCountry` ?

ame-test: From a testing perspective, no, as
* the`src/app/garfile/responsibleperson/get.controller.js` and `src/app/garfile/responsibleperson/index.njk` use `responsibleCounty` instead of `responsibleCountry` (see attached).
* However, I am happy to make the change in the `data-access-api` and elsewhere if wanted.
  - `responsibleCounty` should be `responsibleCountry`.

ho-henry-senior: ah I see, it's everywhere.. ! 😱 does it go as far as the database?

ho-henry-senior (approved):

PR #348, opened 2023-08-31
Feature/fair 92

JHoldergov:

**Definition of Done**


ame-test:
`User.js` was removed as it was not imported anywhere.
* In addition, it was not added to the table (see image).
* [From the documentation](https://sequelize.org/docs/v6/core-concepts/model-basics/#table-name-inference) this is certain, as a `Users` (case sensitive) would be created along side `users` table, but it is not.

ame-test:
Screenshot 2023-09-28 at 10 38 21

ame-test:
Removed `universal-analytics` dependency.
- It was not used within the application, given it was called once to create a variable which was not used (the `visitor` variable).
- From the [documentation](https://github.com/peaksandpies/universal-analytics#universal-analytics), the `visitor` variable is used to create routes from which analytics are recorded. 
  - If the `visitor` variable is not being used, universal analytics is not being used either.

ho-henry-senior (approved):

PR #351, opened 2023-09-07
Feature/fair 85

JHoldergov:

**Definition of Done**

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -52,45 +52,28 @@ steps:
    - push
    - pull_request
 
-#- name: Unit Tests and SonarQube Scanner
-#  pull: if-not-exists
-#  image: quay.io/ukhomeofficedigital/sonar-scanner-node:9fed2b90853321ac9a9c3da548a180f22c5dc4cb
-#  commands:
-#  - export NOTIFY_API_KEY=dummy
-#  - cd src
-#  - npm install
-#  - npm run test-drone
-#  - cd ..
-#  - sonar-scanner -Dsonar.projectVersion=$${DRONE_BUILD_NUMBER}
-#  settings:
-#    group: build-stage
-#  when:
-#    event:
-#    - push
-#    - pull_request
 
-- name: scan-newly-built-image
-  image: 340268328991.dkr.ecr.eu-west-2.amazonaws.com/acp/anchore-submission:latest
+- name: scan-image
   pull: always
-  failure: ignore
-  environment:
+  image: 340268328991.dkr.ecr.eu-west-2.amazonaws.com/acp/trivy/client:latest
+  resources: 
+    limits:
+      cpu: 1000
+      memory: 1024Mi
+  environment: 
     IMAGE_NAME: quay.io/ukhomeofficedigital/egar-public-site-ui:${DRONE_COMMIT}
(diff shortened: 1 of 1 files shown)


JHoldergov:
simple change from Anchore to Trivy as directed by ACP

.drone.yml
```diff
@@ -52,45 +52,26 @@ steps:
    - push
    - pull_request
 
-#- name: Unit Tests and SonarQube Scanner
```
ame-test, line 55: Is the `Unit Tests and SonarQube Scanner` step commented out for a reason? e.g.:
- It was erroring when running the trivy scanner.

JHoldergov: THis was to clean out the pipeline, do we still need it hushed out for later use?

JHoldergov: ps: I didn't comment them out, I can leave them commented out if they are needed for later use,

ame-test: No, I thought the commented code was added in, but I did not see that it was deleted. 🤦 
- Please ignore my comment. It is my mistake.

.drone.yml
```diff
@@ -52,45 +52,26 @@ steps:
...
-  failure: ignore
-  environment:
+  image: 340268328991.dkr.ecr.eu-west-2.amazonaws.com/acp/trivy/client:latest
+  resources: 
+    limits:
+      cpu: 1000
+      memory: 1024Mi
+  environment: 
     IMAGE_NAME: quay.io/ukhomeofficedigital/egar-public-site-ui:${DRONE_COMMIT}
     TOLERATE: high
```
ame-test, line 64: Is the `TOLERATE` environment variable used in Trivy?
* From the docs, it does not seem to be the case (https://docs.acp.homeoffice.gov.uk/how-to/security/trivy/).

.drone.yml
```diff
@@ -52,45 +52,26 @@ steps:
...
-#  when:
-#    event:
-#    - push
-#    - pull_request
-
-- name: scan-newly-built-image
-  image: 340268328991.dkr.ecr.eu-west-2.amazonaws.com/acp/anchore-submission:latest
+- name: scan-image
   pull: always
-  failure: ignore
```
ame-test, line 75: Question: should the `failure: ignore` be removed?
- I am in favour of it being removed, since it means we are likely to fix security incidents as they come.
- Would there be any reason not to e.g., persistent but irrevelant security issues at the image level?

.drone.yml
JHoldergov: the TOLERATE env_var is done away with

.drone.yml
JHoldergov: good point, I agree, the image build is easy to sort it's a matter of triggering a building and the image updates

JHoldergov: At image level we should always try to fix as we go as it does not take much time, I will put up docs regarding fixing the images which is easy to do

ame-test: Ok, that is cool, that clarifies.
* I am happy for a docs update, but it is not required.
* Unless there is an issue I am missing, I am happy for it to be merged.

ame-test (approved):
Looks goods, all comments have been answered.

PR #355, opened 2023-09-14
sGAR ~ 2023.09.14 ~ all departure route has been fixed when submitting to the api

ame-test:
rture for the confirmation route requires a blank [] --> it can not be equal to None
 -->as it uses that [] to go through exceptiojn ids. Therefore the post.contorller.
enforces this. it would be unedefiend, not null as it is a property which would not exist, hence, undefined.

- confirmation route needs the following payload if all the gars are going to departure and not be exceptioned:
```json
{ "exceptions": [] }
```

It was sending something like the below, which caused an error:
```json
{ "exceptions": null }
```

**Definition of Done**

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,7 +3,7 @@ version: '3'
 services:
 
   database:
-    image: postgres:10.4
+    image: postgres:14
     container_name: database
     env_file:
       - .env.dev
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -3,20 +3,24 @@ version: '3'
...
-    image: postgres:10.4
+    image: postgres:14
     container_name: database
     env_file:
       - .env.dev
     ports:
       - "5432:5432"
 
   api:
-    image: api
```
sjose012, line 14: Do we need to remove this line?

sjose012 (approved):

docker-compose.yml
ame-test: We don't need to remove:
- It is from my personal setup of `docker-compose.yml`.
- Useful for getting instant updates from the `api`.
- I will remove it, since it would effect other developers builds.

PR #359, opened 2023-09-25
Feature/country code early release

ame-test:

**Definition of Done**


ame-test:
Sibling pull request: https://gitlab.digital.homeoffice.gov.uk/egar/data-access-api/-/merge_requests/192

## Changes
* Country code value is provided in the backend instead of country code free text now as the responseCountryLabel.
* Responsible person displays the responsibleCountryLabel text instead of the  responsible Country --> which would display the country code.

## Description

* From the UPT changes, the responsible person field changes from being the country free text to a country code.
* If this is released with the UPT changes, then the old gars can potentially cause errors when being sent to AMG (following the amg-test-harness behaviour).
* Therefore, by releasing the country code changes early, the responsible_person gars will all be using country codes, therefore there will be less error during the UPT release.

ame-test:
The above has been made simpler:
* It is now all frontend process.
* The separate `responsibleCountryLabel` has been kept:
   - This is because `responsibleCountry` is being sent a a country code, it is kept a countryCode when it is gotten.
   - `responsibleCountryLabel` is the country label --> it is not the country code and never will be.
   - Therefore, the values are consistent e.g. one will never use the `responsibleCountry` and expect it to be the country label.

ho-henry-senior (approved):

PR #360, opened 2023-09-26
FAIR-120 (Error is displayed while switching from upload file to craf…

sjose012:
…t details)

**Definition of Done**

### DIFF
--- a/src/app/garfile/submit/success/index.njk
+++ b/src/app/garfile/submit/success/index.njk
@@ -13,7 +13,6 @@
 {% block beforeContent %}
 
   {% include "phase.njk" %}
-  {% include "common/templates/includes/back.njk" %}
 
 {% endblock %}
 
--- a/src/app/garfile/supportingdocuments/get.controller.js
+++ b/src/app/garfile/supportingdocuments/get.controller.js
@@ -22,6 +22,7 @@ module.exports = (req, res) => {
       switch (req.query.query) {
         case 'e': // Error returned
           error = [req.session.errMsg];
+          delete req.session.errMsg;
           break;
         case 'v': // Virus detected
           error = [{ message: 'File cannot be uploaded. The file has a virus' }];


src/app/garfile/submit/success/index.njk
```diff
@@ -13,7 +13,6 @@
 {% block beforeContent %}
 
   {% include "phase.njk" %}
-  {% include "common/templates/includes/back.njk" %}
```
ame-test, line 16: Question: Should it be done also for the `app/garfile/submit/failure/index.njk`?
- That currently sets the gar's status to Submitted.
- Therefore, with the back button would send the user to the review page of a "Submitted" (but failed) gar.

src/app/garfile/supportingdocuments/get.controller.js
```diff
@@ -22,6 +22,7 @@ module.exports = (req, res) => {
       switch (req.query.query) {
         case 'e': // Error returned
           error = [req.session.errMsg];
+          delete req.session.errMsg;
```
ame-test, line 25: Question: What issue was caused by this change and how does it fix this?
- This question is out of curiosity.
- I presume that by deleting the property, it does not display the supporting document's `req.session.errMsg` message else where e.g. another form like custom section or responsible people.
- However, I wanted to check so I could understand it better.

ame-test (approved):
The change is small and does not seem to have any bugs, so I am happy for it to be merged.
- I have 2 questions, but the change itself is not flawed.

src/app/garfile/submit/success/index.njk
sjose012: Not sure, what will be the state of the in case of a failure, if it is NOT in the Submitted state, user need an option to go back to the previous page to review and correct the error.

src/app/garfile/supportingdocuments/get.controller.js
sjose012: In one of the scenarios while trying to upload a file to the an already submitted GAR, the error message displayed stays if click on "Aircraft" tab. I think that is because the error message stays in the session. As it is cleared in other parts of the code, I thought of clearing it not to display the error message if moved to other pages.

src/app/garfile/submit/success/index.njk
ame-test: Ok, that is fine.
I thought that when the application submitted a GAR to CBP but failed, it still changed the GAR's status to a Submitted state.
* However, if I am incorrect, then the above is the correct behaviour.

PR #373, opened 2023-10-12
fair-136 - remove departure confirmation screens

ho-henry-senior:
removed amg/departure, amg/departurestatus, removed code that linked to those pages, removed from router


anushasivakumar (commented):
Hi Henry, This looks good to me. This file - src/common/models/Manifest.class.js seems to have some reference to amgDepartureResponseCode. Is this relevant to the changes you have reverted? Thanks!!

ho-henry-senior:
> Hi Henry, This looks good to me. This file - src/common/models/Manifest.class.js seems to have some reference to amgDepartureResponseCode. Is this relevant to the changes you have reverted? Thanks!!

I think this should remain until/unless we remove that field from the data-access.
I think maybe we are best off doing that in a separate pass. Will be removing schema, json serialisation etc
@ame-test - does that sound right? for as long as those fields are present in the json returned by api, the code Anusah it talking about should remain?

ame-test:
> > Hi Henry, This looks good to me. This file - src/common/models/Manifest.class.js seems to have some reference to amgDepartureResponseCode. Is this relevant to the changes you have reverted? Thanks!!
> 
> I think this should remain until/unless we remove that field from the data-access. I think maybe we are best off doing that in a separate pass. Will be removing schema, json serialisation etc @ame-test - does that sound right? for as long as those fields are present in the json returned by api, the code Anusha is talking about should remain?

Yep, it should remain. 
* The `/gar/<string:gar_id>/people` will return thsoe properties are they specificied in the serialisers.
* If one removes them, it should cause an error.
* It should removed when the serialisers are removed.

anushasivakumar (approved):
Thanks both!!

PR #379, opened 2023-10-16
Feature/fair-98 ClamAV config changes

gervy-mbuyi-ho:
We are now using a separate deployment running ClamAV hence the removal of the 2 containers within pod.


ame-test (approved):
Looks good to me.

PR #389, opened 2023-11-02
Feature/master feature node 18

ame-test:

- node 18 upgrade.
- remove the unnecessary grunt file.
- upgrade dependencies.
- Got rid of alter constraint, does nto make sense.

**Definition of Done**


src/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Electronic Gar Site",
   "main": "index.js",
   "scripts": {
-    "dev": "./node_modules/nodemon/bin/nodemon.js start.js",
```
ho-henry-senior, line 7: have you tested if this works when running in a container? 
I can't remember why, but I think the full path was necessary

ame-test: In terms of this one, I think it si fine with and without the local node_modules:
I have tested both, and they both run.
node_modules would be installed locally.
src folder is not copied, but specific folders, so the local node_modules should not be added into the built imge.

ho-henry-senior (approved):

PR #406, opened 2023-12-11
Feature/fair 311 global refactor, handle 1z at the top rather than the bottom

ame-test:

Pull request does the orders `1Z` to the top of the table for all times the garpeople are gotten for a user.
* Thsi is because if a person wanted to edit a do not board person, they can edit them at teh top rather than have to scroll down to find them.
* It only triggers when the checkin message is `1Z`.
* It is on the frontend and not the backend, since this respects the data-access-api design of being only data and no business logic. I can see how it is disagree to this, especially since it causes a disconnected between what the data-access-api returns on a route and what the frontend end does. If one wants to ignore this rule I understand.

I did consider adding scrolling since it may be a better experience on mobile. There are [accessibility concerns](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#accessibility_concerns) though, which may be addressable.

**Definition of Done**


ame-test:
Screenshot 2023-12-11 at 17 41 29
Screenshot 2023-12-11 at 17 41 47
Screenshot 2023-12-11 at 17 41 51
Screenshot 2023-12-11 at 17 42 00
Screenshot 2023-12-11 at 17 42 06

ho-henry-senior (approved):

PR #417, opened 2023-12-19
removed first eventbrite event from homepage

ho-henry-senior:
removes first eventbrite event from homepage


anushasivakumar (approved):
Looks good, thanks

PR #421, opened 2024-01-03
Feature/fair 323 dropdown

ame-test:
Mistake from FAIR 323, adds dropdown for help and guidance page pre user signed in.
- Helps with mobile.
- change is here `src/common/templates/includes/header-pre-user.njk`.

Screenshot 2024-01-03 at 11 17 08
Screenshot 2024-01-03 at 11 17 11


sjose012 (approved):
Looks good

PR #434, opened 2024-01-23
Feature/fair 387 - fixes navigation tests by adding in a check for bad/server error requests

ame-test:
Fixes bug found in by e2e tests (merge request for that change), where :
* if you fetch a gar that does not exist, it is suppose to redirect the user to the dashboard.
* However, currently, it does not.
* Hence, if the gar id is not found, it leads to a filter method being called on it, which causes a TypeError.

The reason this occurs is that the UPT code added in the api functions causes an error if the body.items is undefined.
* Add logging for all endpoints regarding  a 400 bad request or other.
* That means that if we do receive a bad request, we will get an error with a greater depth.

**Definition of Done**

### DIFF
--- a/src/common/services/garApi.js
+++ b/src/common/services/garApi.js
@@ -4,6 +4,16 @@ const endpoints = require('../config/endpoints');
 const autocompleteUtil = require('../utils/autocomplete');
 const travelPermissionCodes = require('../utils/travel_permission_codes.json');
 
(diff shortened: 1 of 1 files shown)


ame-test:
What is on the screen:
Screenshot 2024-01-23 at 10 39 50
Error:
Screenshot 2024-01-23 at 10 40 02

sjose012 (approved):

PR #435, opened 2024-01-24
Feature/fair 400 - add certificate configuration into the egar-public-site-ui

ame-test:
* Add permanent certificate file to the application, meaning that certificate should be consistent across environments.
* Also makes the `base_url` (internal urls) and `base_url_service` (external urls) clearer.

I will delete this [documentation though](https://collaboration.homeoffice.gov.uk/display/BF/sGAR%3Ainvalid+certificate+issue+-+possible+fix).
* Use [ACP docs instead](https://docs.acp.homeoffice.gov.uk/how-to/security/cert-manager/).

**Definition of Done**

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -140,7 +140,7 @@ steps:
       - export TAGGED_VERSION=$${DRONE_COMMIT}
       - export KUBE_SERVER=$${TEST_KUBE_SERVER}
       - export KUBE_TOKEN=$${TEST_KUBE_TOKEN}
-      - export PUBLIC_SITE_EXTERNAL_URL=$${TEST_URL_SERVICE}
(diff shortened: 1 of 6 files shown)


ame-test:
It works in dev, see image, should be ok to go:
Screenshot 2024-01-30 at 13 47 50
Screenshot 2024-01-30 at 13 47 56

ho-henry-senior (approved):
Hey Al, this is approved. 
I think there is quite a lot we might want to change about the deploy process and secret configuration etc so I will probably put some time into planning what can be done.
The goals for this work would be:
- remove anything that is not needed
- make secret management as easy as possible (probably as little in Drone as possible)
- make the environments as similar as possible etc 
I think this change goes some way towards that, but I'd like to do work like this in the future as part of a larger plan.

PR #444, opened 2024-02-16
Feature/fair 408 departure ports are prevented from being too long

ame-test:

This change prevents the departure and arrival port from not being too long when submitting excel or non excel.
- adds validation.
- prevents common whitespace error via trimming (do confirm this).

Pull requests:
* https://github.com/UKHomeOffice/egar-public-site-ui/pull/444
* https://gitlab.digital.homeoffice.gov.uk/egar/data-access-api/-/merge_requests/251
* https://gitlab.digital.homeoffice.gov.uk/egar/e2e-test/-/merge_requests/230

**Definition of Done**

### DIFF
--- a/src/common/utils/autocomplete.js
+++ b/src/common/utils/autocomplete.js
@@ -32,9 +32,31 @@ function getCountryFromCode(countryCode) {
   return countryFromCode === undefined ? countryCode : countryFromCode.label;
 }
 
+/**
+ * generate list of airport codes from the airport list
+ * @param {Array<Object>}
+ * @returns {Array<String>} List of airport codes
+ */
+function listOfAirportCodes(airportList) {
+  const airportCodes = [];
+
+  airportList.forEach((airport) => {
+    const iataCode = airport.id
+    const icaoCode = airport.id2;
+    airportCodes.push(iataCode);
+    airportCodes.push(icaoCode);
+  });
+
+  return airportCodes
+}
+
+const airportCodeList = listOfAirportCodes(airportList);
+
+
 module.exports = {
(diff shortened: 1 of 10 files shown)


src/common/utils/autocomplete.js
```diff
@@ -32,9 +32,31 @@ function getCountryFromCode(countryCode) {
   return countryFromCode === undefined ? countryCode : countryFromCode.label;
 }
 
+/**
+ * generate list of airport codes from the airport list
+ * @param {Array<Object>}
+ * @returns {Array<String>} List of airport codes
+ */
+function listOfAirportCodes(airportList) {
```
sjose012, line 40: When this listOfAirportCodes method will be called? Is it called while the application loads?

sjose012 (approved):

src/common/utils/autocomplete.js
ame-test: It is only execute once, when the `autocomplete.js` module is first required.
* after that is not used --> the [require cache](https://nodejs.org/api/modules.html#:~:text=require.cache%20%23&text=Modules%20are%20cached%20in%20this,will%20result%20in%20an%20error.) handles all usage after that I think.
* The below show it being triggered twice, this is because of nodemon [unloads](https://stackoverflow.com/questions/6676612/unloading-code-modules#:~:text=%22Unloading%22%20a%20node%20module%20means,was%20requiring%20%22%2Flib%2F.) from the node module require cache, the `start.js` module, which triggers the autocomplete module again.

Screenshot 2024-02-16 at 13 08 31

PR #446, opened 2024-02-22
Validation added for zero month entries

justinhyltonHO:
h3. {color:#0847a6}Background{color}
Whist working on ticket FAIR-172 on the GAR Departure details form.  I observed the following when I entered "0" for day, month or year in the departure date input fields and submited the form:
# The {color:#e21b29}"Departure date must be today and cannot be more than 1 month in the future"{color} error message is displayed twice.
# Also, the error message {color:#e21b29}"Enter a real departure date"{color} is not displayed when a "0" month is submitted.  See attachment for an example.

h3. {color:#0847a6}Affected Screens{color}
# GAR Arrival details
# GAR Departure details

**Definition of Done**
- [ ☑️] Have you built the code locally and confirmed its working?
- [☑️] ] Is the build confirmed to be passing on CI?
- [ ☑️]] Have you added enough relevant unit tests for your change?
- [ ] ] Has this change been tested against the Acceptance Criteria?


src/common/utils/validator.js
```diff
@@ -91,16 +91,16 @@ function valuetrue(value) {
...
-    case 1:
-      return (y % 4 === 0 && y % 100) || y % 400 === 0 ? 29 : 28;
-    case 8:
-    case 3:
-    case 5:
-    case 10:
-      return 30;
+  switch (m) {
+    case '2':
+      return (y % 4 == 0 && y % 100 != 0) || (y % 100 == 0 && y % 400 == 0) ? '29' : '28';
```
ame-test, line 96: Question: is there a good reason to use this code over using the native javascript e.g. `new Date(y,m + 1, 0).getDate()` (https://stackoverflow.com/questions/1184334/get-number-days-in-a-specified-month-using-javascript) or an npm library e.g. [luxon](https://www.npmjs.com/package/luxon?activeTab=dependencies), etc, etc?
- Dates and time can be notorious complicated.
- Could contain a bug from edge case.

Though
- Not downloading a library or using edge case behaviour is nice.

src/common/utils/validator.js
```diff
@@ -117,17 +117,23 @@ function isPrintable(value) {
...
   if (isNumeric(m)) {
-    return m >= 0 && m <= 12;
+    return m >= 1 && m <= 12;
+  }
+  return false;
+}
+
+function validYear(y) {
+  if (isNumeric(y)) {
+    return /^[1-9][0-9]{3}?$/gm.test(y); // Matches 4 digit years between 1000 and 9999
```
ame-test, line 136: Question: should regex be here or not?
* e.g. something like `1000 >= y && y <= 9999` does is what the comment is saying.

This is asked as the code is looking for a particular valid year number, using regex seems like overkill.

src/common/utils/validator.js
```diff
@@ -212,6 +218,14 @@ function currentOrPastDate(dObj) {
     return true;
   }
 
+  /*
+    Returns true if supplied dates are invalid to avoid duplicate error messages being displayed.
+    The dateNotTooFarInFuture function will cover this error.
+  */
+  if(validDay(dObj.d, dObj.m, dObj.y) == false || validMonth(dObj.m) == false || validYear(dObj.y) == false){
```
ame-test, line 225: Please stop using `==` or `!=` in javascript code.
* By removing strict type checking from comparsion, it remove information of what `validDay`, `validMonth` or `validYear` return e.g. is it returning the year number (e.g. `2024`) or `false` boolean.

It is not a issue here, but if done in other parts of the code, it can lead to hard to find bugs.
* There does not seem be a good reason to use loose equality here.

ame-test (approved):
The code is ok.
* Questions are minor and optional, the code seems to work.

src/common/utils/validator.js
ame-test: Has been done, thanks for doing so.

src/common/utils/validator.js
ame-test: Has been done, thanks for doing so.

src/common/utils/validator.js
ame-test: Has been done, thanks for doing so.

Note, the `validYear` will only return `true` for number strings (`validYear("2000")` or `validYear("3010")`) rather than actual numbers (`validYear(2000)` or `year(3010)`).
* Is this intended behaviour or not?
* It should not error in this case, it can remain in if it is desired - if possible, give the reason why.

Part of code I am referring to:
```javascript
function validYear(y) {
  if (isNumeric(y)) {
    return (y.length === 4) && (y >= 1000 && y <= 9999);
  }
  return false;
}
```

Reasoning:
* This is because `Number` primitive in javascript (`2000` rather than `"2000"`) **don't** have a `length` property.
* `String` primitives e.g. `"bananas on toast"` and `"2000"` do have a `length` property.
* Therefore, even though `isNumeric(2000)` returns `true`, the below check will return `false`.
* If I am wrong, let me know.
* Since currently the app seems to send string number year in, so it is fine. Concern  is it may change input to a number year in the future though.

ame-test:
Hi @justinhyltonHO , thank you for making the above changes.

Notes on potential future bug in `validYear`, but I can accept an urgent merge.
* https://github.com/UKHomeOffice/egar-public-site-ui/pull/446#discussion_r1501863148

Making comment on Sunday at 17:28 as I missed 10 minutes on Friday due various reasons. I am making it up back today.

PR #450, opened 2024-03-04
Feature/fair 436

ho-henry-senior:
- Responsible person email address is required
- Label for responsible person email no longer says 'optional'
- Unit test that checks validations fail when responsible person email is absent
- (Other unit test brought in line with responsible person email being present)

image


ame-test (approved):
Looks good to me.

PR #452, opened 2024-03-11
Feature/fair 437

ho-henry-senior:
- Removed email send when GAR is submitted (this is now done by data-integr-cbp)
- Updated GAR submit success page to exclude GAR ID (we are now using the CBP reference ID which is not available when that page is displayed).
- Updated Manage GARs page to display CBP Id (when available) in place of GAR Id.
- A couple of misc changes where I'd noticed errors.

### DIFF
--- a/src/app/garfile/submit/success/index.njk
+++ b/src/app/garfile/submit/success/index.njk
@@ -7,7 +7,7 @@
 {% endblock %}
 
 {% block header %}
-{% include "header-user.njk" %}
+  {% include "header-user.njk" %}
 {% endblock %}
 
 {% block beforeContent %}
@@ -18,7 +18,6 @@
 
 {% block content %}
 
-
   {% import "common/templates/includes/macros.njk" as m %}
 
   {% if errors %}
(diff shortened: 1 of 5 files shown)


src/app/garfile/submit/success/index.njk
```diff
@@ -18,50 +18,34 @@
...
-        </strong>
-      </div>
+      </h1>
 
-       <h2 class="govuk-heading-m">
-         What happens next
-       </h2>
-
-       <p class="govuk-body-m">
+      <p class="govuk-body-m">
```
ame-test, line 33: Minor: From [FAIR 437](https://collaboration.homeoffice.gov.uk/jira/browse/FAIR-437), the text has poor contrast with dark blue-green color. Could be an accessibility problem or not?

ame-test (approved):
Looks good.

I have an accessibility comment, but I trust your judgement on whether to handle it or not.

Otherwise, please merge.

src/app/garfile/submit/success/index.njk
ho-henry-senior: This passes WCAG AA so I think OK for the moment.

PR #453, opened 2024-03-12
added endpoint to export organisation users as csv

ho-henry-senior:
- Added endpoint that exports people as CSV
- Added link/button to people homepage

image

[people-2024-03-12.csv](https://github.com/UKHomeOffice/egar-public-site-ui/files/14574641/people-2024-03-12.csv)

----------------

the below tasks were actually not requested and implemented by mistake, but I have left is as they seem potentially useful.

- Added endpoint that exports organisation users as CSV
- Added link/button to organisation homepage
- (fixed table layout in organisation homepage when no action link is visible)

image

[Henry-Organisation-users-2024-03-12.csv](https://github.com/UKHomeOffice/egar-public-site-ui/files/14573274/Henry-Organisation-users-2024-03-12.csv)

### DIFF
--- a/src/app/organisation/exportusers/get.controller.js
+++ b/src/app/organisation/exportusers/get.controller.js
@@ -23,7 +23,6 @@ const writeUsersAsCSVtoResponse = (res, orgUsers, orgName) => {
 module.exports = (req, res) => {
     logger.debug('In organisation export users get controller');
     const cookie = new CookieModel(req);
-    const errMSg = { message: 'Failed to get orgusers' };
 
     orgApi.getUsers(cookie.getOrganisationId())
         .then((values) => {
--- a/src/app/organisation/exportusers/get.controller.js
+++ b/src/app/organisation/exportusers/get.controller.js
@@ -12,6 +12,7 @@ const writeUsersAsCSVtoResponse = (res, orgUsers, orgName) => {
 
     const orgSlug = orgName.replace(/[^A-z]/, '-')
     const dateSlug = new Date().toISOString().slice(0,10);
+
     res.setHeader('Content-disposition', 'attachment; filename=' + orgSlug + '-users-' + dateSlug + '.csv');
     res.setHeader('Content-Type', 'text/csv');
     
@@ -26,6 +27,7 @@ module.exports = (req, res) => {
 
     orgApi.getUsers(cookie.getOrganisationId())
         .then((values) => {
+            
             const orgUsers = JSON.parse(values).items.map((orgUser) => {
 
                 return [orgUser.userId, orgUser.firstName, orgUser.lastName, orgUser.email, orgUser.role.name, orgUser.state];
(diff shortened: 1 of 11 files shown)


src/app/people/export/get.controller.js
```diff
@@ -0,0 +1,35 @@
...
+    const dateSlug = new Date().toISOString().slice(0, 10);
+    res.setHeader('Content-disposition', 'attachment; filename=people-' + dateSlug + '.csv');
+    res.setHeader('Content-Type', 'text/csv');
+    res.write(csvStringifier.getHeaderString());
+    res.write(csvStringifier.stringifyRecords(people));
+    res.end();
+}
+
+module.exports = (req, res) => {
+    logger.debug('In people export get controller');
```
anushasivakumar, line 21: I think it may be useful to log the id of user who is downloading the data, to keep a record for security/auditing purposes - as there is PII being downloaded/viewed here

src/app/people/export/get.controller.js
```diff
@@ -0,0 +1,35 @@
...
+    const cookie = new CookieModel(req);
+
+    personApi.getPeople(cookie.getUserDbId(), 'individual')
+        .then((values) => {
```
anushasivakumar, line 25: What happens here if there is an empty string?

src/app/organisation/exportusers/get.controller.js
```diff
@@ -0,0 +1,38 @@
...
+    
+    res.write(csvStringifier.getHeaderString());
+    res.write(csvStringifier.stringifyRecords(orgUsers));
+    res.end();
+}
+
+module.exports = (req, res) => {
+    logger.debug('In organisation export users get controller');
+    const cookie = new CookieModel(req);
+    const errMSg = { message: 'Failed to get orgusers' };
```
anushasivakumar, line 26: is errMsg used here? or may be we need to add error handling?

src/app/organisation/exportusers/get.controller.js
```diff
@@ -0,0 +1,38 @@
...
+
+    orgApi.getUsers(cookie.getOrganisationId())
```
anushasivakumar, line 28: Do you think adding any tests for these new modules would be helpful? Also perhaps we can create any e2e tests if that is relevant..

anushasivakumar (commented):
Thanks Henry, changes looks good to me. Added some thoughts to the PR.

src/app/people/export/get.controller.js
ho-henry-senior: I updated the logging to include userId.

src/app/people/export/get.controller.js
ho-henry-senior: if there are no people the api returns a json empty array [] and the export is just the headers.

src/app/organisation/exportusers/get.controller.js
ho-henry-senior: errMsg was not used, have removed. we could add error handling, I'm not sure what we'd do given it's a download process rather than a page that is rendered to user. if if there are exceptions they will be logged?

src/app/organisation/exportusers/get.controller.js
ho-henry-senior: I added a couple of unit tests for this.

anushasivakumar:
Thanks for making the changes, Approved.

anushasivakumar (approved):

PR #458, opened 2024-03-26
removed GAR support phone number from help page

ho-henry-senior:
removed support phone number from help page


ho-henry-senior:
I left the commented html in in case we need to add a phone number. Could remove at a later date.

ame-test (approved):
Looks good.

PR #492, opened 2024-05-08
sGAR ~ 2024.05.08 ~ event added.

ame-test:

event added.

Screenshot 2024-05-08 at 11 29 26

**Definition of Done**

### DIFF
--- a/src/app/welcome/index.njk
+++ b/src/app/welcome/index.njk
@@ -116,7 +116,7 @@ Create GAR | Start
 				<section class="govuk-tabs__panel govuk-tabs__panel--hidden" id="events">
 					<h3 class="govuk-heading-s">Home Office Engagement Events</h3>
 					<p>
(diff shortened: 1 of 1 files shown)


src/app/welcome/index.njk
```diff
@@ -129,11 +129,12 @@ Create GAR | Start
 						To register for an event, please click on a link below:
 					</p>
 					<ul class="govuk-list govuk-list--bullet">
+						{#<li>No upcoming events.</li>#}
```
justinhyltonHO, line 132: Minor Change:  I think it is better to add a space before and after the list tags for readability.

justinhyltonHO (approved):

src/app/welcome/index.njk
ame-test: I will live with the current formatting, since we should get a linter sound, which will determine the code styles.

PR #499, opened 2024-05-20
sGAR ~ 2024.05.19 ~ other validations are incorrect

ame-test:

Other document type had separate validations to the people sections for seemingly no reason.

**Definition of Done**

### DIFF
--- a/src/test/people/edit/post.controller.test.js
+++ b/src/test/people/edit/post.controller.test.js
@@ -166,7 +166,8 @@ describe('Person Edit Post Controller', () => {
       expect(res.render).to.not.have.been.called;
     });
   });
-  it('should render with people other document type validating the alpha numeric stuff', () => {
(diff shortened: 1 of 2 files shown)


src/test/people/edit/post.controller.test.js
```diff
@@ -166,4 +166,31 @@ describe('Person Edit Post Controller', () => {
       expect(res.render).to.not.have.been.called;
     });
   });
+  it('should render with people other document type validating the alpha numeric stuff', () => {
```
justinhyltonHO, line 169: I think you could make the description clearer.  Something like:  'should render with validation messages if "Other" document type is selected and special characters, apostrophes or numbers are present in the given name'

I presume the req.body.firstName is "Given Name" on the form.  Maybe you could leave a comment to indicate this.

justinhyltonHO (approved):

PR #507, opened 2024-06-13
Feature/fair 526

ame-test:

FAIR 526:
- this changes the cancellation email section so that if the user:
 - Has not submitted a gar, they get no cancellation email.
 - If they have submitted a GAR, they do a get a cancellation, they get submission reference.

Previously it was sending a email for all GARs, with the gar id as the reference.
- this was fine when the gar id was the reference for all users, but since then it has been replaced withthe CBP id.
- Without the GAR id, non gar emails beign cancelled,t he gar id provides no context teh user on which gar was cancelled on to the dev team for support.
- If a GAR is getting cancelled before the submission, it's cancellation does not do anything, it is the same as if the user did not create the gar int eh frist.
- - however, submitted joruneys beign cancelled do have a meaning, a journey that was happening is no happening, hence the confirmination is needed here, so the user knwo the journey is cancelled.

There are question on amend process e.g. if you cancel an amend GAR, it will cancel the previous journey, which could cause a hard to find bug where a user cancels an ammend because teh flight is ok, but has cancelled the journey, which causes issues when they go fly.

**Definition of Done**


ame-test:
Tests:
image

ame-test:
updates email, gov ntoify and example:
Screenshot 2024-06-13 at 13 17 23
Screenshot 2024-06-13 at 13 17 57

ame-test:
airport validation being added is being it was mistakenly removed fromt he file, whcih causes an udnefined variable error.

src/app/garfile/view/get.controller.js
```diff
@@ -2,6 +2,7 @@ const CookieModel = require('../../../common/models/Cookie.class');
 const logger = require('../../../common/utils/logger')(__filename);
 const garApi = require('../../../common/services/garApi');
 const manifestFields = require('../../../common/seeddata/gar_manifest_fields.json');
+const airportValidation = require('../../../common/utils/airportValidation');
```
ninaknowlesHO, line 5: Why did you need to add this in?

src/test/garfile/view/post.controller.test.js
```diff
@@ -403,7 +403,7 @@ describe('GAR view post controller', () => {
     };
 
     callController().then().then(() => {
-      expect(garApiGetStub).to.have.been.calledOnceWithExactly('GAR-ID-EXAMPLE-1');
+      expect(garApiGetStub).to.have.been.calledOnceWithExactly('GAR-ID-EXAMPLE-1', true);
```
ninaknowlesHO, line 406: What happens when it is false?

ninaknowlesHO (approved):

src/app/garfile/view/get.controller.js
ame-test: From discussion, it was that opensearch raised airport validaiton error, get controller did not have an airport validaiton reference, thus I added it.
* Seemed to be trigger on the amend get.controller.

PR #511, opened 2024-06-24
Other minor testing bugs picked up --> more general

ame-test:

These are 3 things found during tests:
* By checking for port code length rather than portChoice in `post.controller.js` for arrival` it will only send the port validation there than the latitude/longitutde validation. Likely the port is `' '`, so an error is raised, but it should use the correct validation.
* editCraft does update to `editCraft` cookie value when saving stuff, instead it is goes to another value. This means that whatever changes the users make, they are nto saved so if an error occurs, their work on the edit is lost. This fixes that.
* I also noticed that teh `/user/viewDetails` psot controller.js is not used anywhere, so it was deleted.

**Definition of Done**


wesleymorelandHO (approved):
Looks good to me. Went through changes on a call, no issues.

sjose012 (approved):

PR #512, opened 2024-06-27
FAIR-524 Temporary Home Page Notice

SaritaMore7:

**Definition of Done**


src/app/welcome/get.controller.js
```diff
@@ -1,6 +1,7 @@
 const logger = require('../../common/utils/logger')(__filename);
+const homePageMessage = process.env.HOMEPAGE_MESSAGE;
```
ho-henry-senior, line 2: Hey Sarita, the other environment variables are pulled in `./src/common/config/index.js` I think it would be good, if possible to do the same with HOMEPAGE_MESSAGE and then import config.

SaritaMore7: done

ho-henry-senior (approved):

PR #516, opened 2024-07-07
FAIR-483 (Manage responsible person)

sjose012:

**Definition of Done**


sjose012:
Screenshot 2024-07-07 at 12 51 26
Screenshot 2024-07-07 at 12 53 07

src/test/resperson/edit/get.controller.test.js
```diff
@@ -0,0 +1,105 @@
...
+        res = {
+            redirect: sinon.spy(),
+            render: sinon.spy(),
+        };
+        sessionSaveStub = sinon.stub(req.session, 'save').callsArg(0);
+        getResPersonStub = sinon.stub(resPersonApi, 'getResPersonDetails');
+    });
+    req = {
+        body: {
+            responsibleGivenName: 'Benjamin',
```
ame-test, line 37: `Optional`: Test in general could use the `fixtures.js` file to store repeats.

src/public/javascripts/resperson_utils.js
```diff
@@ -0,0 +1,66 @@
+$(document).ready(function () {
+    $('th').each(function (col) {
+        $(this).hover(function () {
+            $(this).addClass('focus');
+        }, function () {
+            $(this).removeClass('focus');
+        });
+        $(this).click(function () {
```
ame-test, line 8: `Question`: Where is the `asc`/`desc` code used?
* I can't find it.

src/app/resperson/get.controller.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+    if (req.session.errMsg) {
+      const { errMsg } = req.session;
+      delete req.session.errMsg;
+      return res.render('app/resperson/index', {
+        cookie, resPersons, errors: [errMsg],
+      });
+    }
+    if(resPersons.message) {
+      logger.info(`Failed to get saved people: ${resPersons.message}`)
```
ame-test, line 26: This `if(resPersons.message) {` is odd:
* As far as I am away `resPersonApi.getResPersons` --> `/<user_id>/responsiblepersons` does nto return delibrate erors inthe `{ "message": xyz }` format.
* If it did, it would cause a minor bug with `src/app/resperson/index.njk` as `resPersons` is an object, objects can have `|` [length](https://mozilla.github.io/nunjucks/templating.html#length).
* The log message is incorrect as well with `Failed to get saved people:`
```
    {% if resPersons | length %}
              {% for resPerson in resPersons %}
 ```

File: src/app/resperson/index.njk [hunk 4]
```diff
@@ -0,0 +1,88 @@
...
+              <th class="govuk-table__header" id="email" scope="col">{{__('field_email')}}</th>
+              <th class="govuk-table__header" scope="col">{{__('table_action')}}</th>
+            </tr>
+          </thead>
+          <tbody class="govuk-table__body" id="resPerson_table_row">
+            {% if resPersons | length %}
+              {% for resPerson in resPersons %}
+                  <tr class="govuk-table__row">
+                    <td class="govuk-table__cell">
+                      <button id="edit-edit_reponsiblePerson-{{ loop.index }}" class="govuk-link btn-link"
```
ame-test, line 56: `Optional`: Larger design change so I understand if this should be consider at a late date or not at all.
* Instead of `editResponsiblePerson` and `value="{{resPerson.responsiblePersonId}}` or `name="deleteResponsiblePerson" value="{{resPerson.responsiblePersonId}}"`, `formaction` property on the button could be used instead to direct the edit or delete actions e.g.
```njk
{# instead of the below #}
<button id="edit-edit_reponsiblePerson-{{ loop.index }}" class="govuk-link btn-link" 
                      title="Edit details for {{ resPerson.responsibleGivenName }} {{ resPerson.responsibleSurname }}" 
                      name="editResponsiblePerson" value="{{resPerson.responsiblePersonId}}">{{resPerson.responsibleGivenName}}
                      </button>
{# This can be done #}
<button id="edit-edit_reponsiblePerson-{{ loop.index }}" class="govuk-link btn-link" 
                      title="Edit details for {{ resPerson.responsibleGivenName }} {{ resPerson.responsibleSurname }}" 
                     formaction="resperson/edit" formmethod="/get" name="editResponsiblePerson" value="{{resPerson.responsiblePersonId}}">{{resPerson.responsibleGivenName}}
                      </button>
```
This removes the need for `src/app/resperson/post.controller.js` routing.

src/common/services/resPersonApi.js
```diff
@@ -0,0 +1,102 @@
+const request = require('request');
+const logger = require('../utils/logger')(__filename);
+const endpoints = require('../config/endpoints');
+
+module.exports = {
+
+    create(userId, resPerson) {
+        logger.debug(JSON.stringify(resPerson));
```
ame-test, line 8: `Question`: Logging this can reveal `JSON.stringify(resPerson)` details e.g. `email address` `phone number` `firstName`, etc, etc.
* Is this an ok risk or not?
* I don't mind either way, as long as the risk has been accounted.

src/app/resperson/index.njk
```diff
@@ -0,0 +1,88 @@
...
+                    <td class="govuk-table__cell">
+                      <button id="edit-edit_reponsiblePerson-{{ loop.index }}" class="govuk-link btn-link" 
+                      title="Edit details for {{ resPerson.responsibleGivenName }} {{ resPerson.responsibleSurname }}" 
+                      name="editResponsiblePerson" value="{{resPerson.responsiblePersonId}}">{{resPerson.responsibleGivenName}}
+                      </button>
+                    </td>
+                    <td class="govuk-table__cell" scope="row">{{resPerson.responsibleSurname}}</td>
+                    <td class="govuk-table__cell">{{resPerson.responsibleEmail}}</td>
+                    <td class="govuk-table__cell ">
+                      <button id="delete_responsible_person_{{ loop.index }}" class="govuk-link btn-link" name="deleteResponsiblePerson" value="{{resPerson.responsiblePersonId}}">{{__('table_delete')}}</button>
```
ame-test, line 64: `Question`: there no confirmation before delete, is this ok for now?
* I am happy with no confirmination, since any more changes seems quite a lot.
* However, it should be mentioned.

src/app/resperson/get.controller.js
```diff
@@ -0,0 +1,38 @@
+const logger = require('../../common/utils/logger')(__filename);
+const CookieModel = require('../../common/models/Cookie.class');
+const resPersonApi = require('../../common/services/resPersonApi');
+
+module.exports = async (req, res) => {
+  logger.debug('In responsible person get controller');
+  const cookie = new CookieModel(req);
+  const errMSg = { message: 'Failed to get saved responsible persons' };
```
ame-test, line 8: `nit`: This variable `errMSg` is too easy to conflate with `errMsg` from the `req.session`.

src/app/resperson/delete/get.controller.js
```diff
@@ -0,0 +1,32 @@
...
+  if (responsiblePersonId === undefined) {
+    res.redirect('/resperson');
+    return;
+  }
+
+  resPersonApi.deleteResponsiblePerson(cookie.getUserDbId(), responsiblePersonId)
+    .then((apiResponse) => {
+      const parsedResponse = JSON.parse(apiResponse);
+      if (Object.prototype.hasOwnProperty.call(parsedResponse, 'message')) {
+        req.session.errMsg = errMsg;
```
ame-test, line 20: No log here if an error occurs, may make it hard to debug in the future.

src/app/garfile/resperson/post.controller.js
```diff
@@ -0,0 +1,67 @@
...
+            logger.error(err);
+            res.render('app/garfile/resperson/index', context, {
+                errors: [{
+                    message: 'Problem retrieving GAR',
+                }],
+            });
+        };
+    } else {
+
+        const responsiblePerson = utils.getResponsiblePersonFromReq(req);
```
ame-test, line 28: `nit`: Mix of the `async` and promises.
* In the future, should choose one or the other (ideally `async`, but it is another choice.
* Changing this can break tests and is large, so I don't really want it now.

ame-test (commented):
There are more comments I can make, but that would take a lot fo time and could lead to bugs being introduced.
* I would prefer to get merged and done rather since the pull requests are really big .


I apologise for the delay by the way, I mentioned I wanted to do it by Tuesday/Wednesday, but it was done on Friday.

src/public/javascripts/resperson_utils.js
sjose012: This code is used to sort when one click on the table header, this is used though out ex. for people, crafts, org users etc.

src/app/resperson/get.controller.js
sjose012: 1. It do returns an error message { "message": xyz } format, it is coming from backend, it is not very obvious but it is coming from @validate_user type.
2. resPersons is not an object here it is an array/list, list of responsible persons for a user
3.  will update the log message.

src/common/services/resPersonApi.js
sjose012: This log is not be needed, removed.

src/app/resperson/index.njk
sjose012: There are no confirmation before delete as of now, this follows the way as it is done for crafts, people etc. Not sure what is the best approach to show a confirmation message, browsers like Brave can disable the popup confirmation message and can create problem.

src/app/resperson/get.controller.js
sjose012: renamed to errMessage to avoid confusion

src/app/resperson/delete/get.controller.js
sjose012: log added.

src/public/javascripts/resperson_utils.js
ame-test: This makes sense, I did not read the code properly, thanks.

src/app/resperson/get.controller.js
ame-test: Point 1 and 3 and are good.

On point 2, I am uncertain. I though that if `resPersons.message` was `true`, then `resPersons` is necessarily an object, otherwise it would not have the `resPersons.message` e.g. below:
```js
    const response = await resPersonApi.getResPersons(cookie.getUserDbId())
    const resPersons = JSON.parse(response); // {'message': 'User not registered'}
    if (req.session.successMsg) {
      const { successMsg, successHeader } = req.session;
      delete req.session.successHeader;
      delete req.session.successMsg;
      return res.render('app/resperson/index', { cookie, resPersons, successMsg, successHeader });
    }
    if (req.session.errMsg) {
      const { errMsg } = req.session;
      delete req.session.errMsg;
      return res.render('app/resperson/index', {
        cookie, resPersons, errors: [errMsg],
      });
    }
    if(resPersons.message) {
      logger.info(`Failed to get saved responsible persons: ${resPersons.message}`)
      return res.render('app/resperson/index', {
        cookie, 
        resPersons, // this would be {'message': 'User not registered'}
        errors: [errMessage],
      });
    }
    return res.render('app/resperson/index', { cookie, resPersons });
    ```

src/common/services/resPersonApi.js
ame-test: This is good.

src/app/resperson/index.njk
ame-test: This is ok.

src/app/resperson/delete/get.controller.js
ame-test: This is good.

src/app/resperson/get.controller.js
ame-test: That is fine.

src/app/resperson/index.njk
sjose012: updated and removed the post controller

src/test/resperson/edit/get.controller.test.js
ame-test: Agreed in stand up not to do in this current pull request.

src/app/resperson/index.njk
ame-test: tests good.

ame-test (approved):
Looks good, please merge at your convenince.

PR #524, opened 2024-08-05
Fair 593 - onbound gar and inbound gars sent dat to apprioprate systems.

ame-test:
- send checkin request for all items when submitting.
- exception send for all checkin messages, not just when checkin response is received.
- async is used due to making the unit tests understandable.

----
Passes e2e tests (local) and a manual test.

**Definition of Done**

### DIFF
--- a/src/app/garfile/cancel/post.controller.js
+++ b/src/app/garfile/cancel/post.controller.js
@@ -7,13 +7,13 @@ const config = require('../../../common/config');
 module.exports = (req, res) => {
   logger.debug('In garfile / cancel post controller');
   const cookie = new CookieModel(req);
-  const isCancelled = true
+  const isCancelled = true;
 
   garApi.getPeople(cookie.getGarId())
     .then((res) => {
       const garpeople = JSON.parse(res).items;
       const passengersToRaiseException = garpeople
-        .filter(garperson => garperson.amgCheckinResponseCode !== null)
+        .filter(garperson => garperson.amgHasDeparted !== null)
         .map(garperson => garperson.garPeopleId);
 
       return passengersToRaiseException;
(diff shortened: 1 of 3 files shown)


src/app/garfile/cancel/post.controller.js
```diff
@@ -7,13 +7,13 @@ const config = require('../../../common/config');
...
   const cookie = new CookieModel(req);
-  const isCancelled = true
+  const isCancelled = true;
 
   garApi.getPeople(cookie.getGarId())
     .then((res) => {
       const garpeople = JSON.parse(res).items;
       const passengersToRaiseException = garpeople
-        .filter(garperson => garperson.amgCheckinResponseCode !== null)
+        .filter(garperson => garperson.amgHasDeparted !== null)
```
sjose012, line 16: what is the benefit of changing the filter condition here?

ame-test: In summary, this ensure if a check in message is sent and the GAR is cancelled --> all checkin message receive a cancellation message.

----
The amgHasDeparted change is to make sure the cancellation message is always sent from if a checkin message has been sent, not if it has been processed.
This should be done since we are going to send more messages:
Previously, we cancelled GAR by checkin response code e.g. 0A
However, there is a chance, that if we are still processing checkin messages and a user cancels, they will cancel the checkin messages which have been processed and any unprocessed messages will remain checked in.
However, if we use amgHasDeparted and always set it true for checkin messages and false for cancellations messages and null if the passenger has not been sent to AMG, we can use that value instead.
Since all messages, regardless of whether they have been processed e.g. have a 0A response, will has amgHasDeparted set to true, those passenger will be cancelled.
They will not be cancelled if the amgHasDeparted value is none.
To ensure the above, the SQS queue should be a FIFO rather than a standard queue, which can send messages out of order /duplicates.

PR #525, opened 2024-08-06
Fair 470 organisation users pagination

SaritaMore7:

**Definition of Done**

### DIFF
--- a/src/app/organisation/exportusers/get.controller.js
+++ b/src/app/organisation/exportusers/get.controller.js
@@ -4,7 +4,6 @@ const logger = require('../../../common/utils/logger')(__filename);
 const createArrayCsvStringifier = require('csv-writer').createArrayCsvStringifier;
 
 
-
 const writeUsersAsCSVtoResponse = (res, orgUsers, orgName) => {
     const csvStringifier = createArrayCsvStringifier({
         header: ['Id', 'First Name', 'Last Name', 'Email', 'Role', 'State']
@@ -15,7 +14,7 @@ const writeUsersAsCSVtoResponse = (res, orgUsers, orgName) => {
 
     res.setHeader('Content-disposition', 'attachment; filename=' + orgSlug + '-users-' + dateSlug + '.csv');
     res.setHeader('Content-Type', 'text/csv');
-    
+
     res.write(csvStringifier.getHeaderString());
     res.write(csvStringifier.stringifyRecords(orgUsers));
     res.end();
@@ -25,9 +24,11 @@ module.exports = (req, res) => {
     logger.debug('In organisation export users get controller');
     const cookie = new CookieModel(req);
 
-    orgApi.getUsers(cookie.getOrganisationId())
+    // to retrieve all rows, the 1st to the last row from the users table
+    // set the offset up to the end of the result set (a large number for the second parameter).
+    orgApi.getUsers(cookie.getOrganisationId(), 1, 999999999999999)
         .then((values) => {
-            
+
             const orgUsers = JSON.parse(values).items.map((orgUser) => {
 
                 return [orgUser.userId, orgUser.firstName, orgUser.lastName, orgUser.email, orgUser.role.name, orgUser.state];
--- a/src/app/organisation/index.njk
+++ b/src/app/organisation/index.njk
@@ -9,20 +9,6 @@
 
 {% block header %}
   {% include "header-user.njk" %}
-  
-  <style>
-    .organisation-user-search-button {
-      background:#00823b;border:none; position: relative; cursor:pointer;
-    }
-    #organisation-search button{
-      height: 20%; width: 20%;
-      background-color: green;
-    }
-    #organisation-search {
-      float: left;
-      margin-bottom: 10px;
-    }
-  </style>
 {% endblock %}
 
 {% set active = '/organisation' %}
@@ -32,6 +18,7 @@
 {% endblock %}
 
 {% block content %}
+  <script type="text/javascript" src="/javascripts/organisation.js"></script>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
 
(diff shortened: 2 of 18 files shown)


src/app/organisation/exportusers/get.controller.js
```diff
@@ -25,9 +24,11 @@ module.exports = (req, res) => {
     logger.debug('In organisation export users get controller');
     const cookie = new CookieModel(req);
 
-    orgApi.getUsers(cookie.getOrganisationId())
+    // to retrieve all rows, the 1st to the last row from the users table
+    // set the offset up to the end of the result set (a large number for the second parameter).
+    orgApi.getUsers(cookie.getOrganisationId(), 1, 999999999999999)
```
sjose012, line 29: Are you retrieving all the users at this line?

SaritaMore7: yes, all the users based on the organisation

src/app/organisation/index.njk
```diff
@@ -65,16 +80,48 @@
             $("#organisation_table_row tr").filter(function () {
```
ame-test, line 80: There are 2 separate script tag (not including the dynamic one).
* They should be in a separate file or at least the same script tag.

src/app/organisation/index.njk
```diff
@@ -65,16 +80,48 @@
...
               $(this).toggle($(this).text().toLowerCase().indexOf(value) > -1)
             });
+            
           });
+        
+          $("#searchUsers").on("click", function () {
```
ame-test, line 86: Question: Why does this code exist e.g. it looks liek it dynmic creates a hidden input value, add it the organisation form and submits that form?
- Why is the input for search not part of the `organisation` form or in a separeate form which points to the same endpoint?

I don't mind the code existing, it is just kinda weird.

src/app/organisation/index.njk
```diff
@@ -1,13 +1,28 @@
...
+{% import "common/templates/includes/pagination.njk" as pagination %}
 
 {% block pageTitle %}
   Manage organisation | Submit a General Aviation Report (GAR)
 {% endblock %}
 
 {% block header %}
   {% include "header-user.njk" %}
+  
+  <style>
```
sjose012, line 13: is there a gov-uk style that we can us?

src/app/organisation/index.njk
```diff
@@ -65,16 +80,48 @@
...
+            });
+      </script>
+      {% endif %}
+
       <form id="organisation" action="/organisation" class="form-group" method="post" role="form" >
+        <div class="govuk-grid-column-two-thirds">    
+          <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--xl govuk-!-margin-top-0">
+          <p class="govuk-body-m">Search for people within the organisation,  {{ cookie.getOrganisationName() }}</p>
+          <input class="govuk-input govuk-input--width-20" id="organisation-search" name="organisation-search" type="text" value="{{ searchUserName }}" placeholder="Search">
+
```
ame-test, line 115: Minor: You can submit the form without any value, which does not make a lot of sense.
* leads to endless load times when the app has items in the form.
* Otherwise, if there are no items, it loads very quickly when search blank

ame-test: Screenshot 2024-08-06 at 15 47 05

src/app/organisation/index.njk
SaritaMore7: done.  removed all js code from organisation nunjukes file and put it in the new separate js file - 'src/public/javascript/organisation.js'

src/app/organisation/index.njk
SaritaMore7: modified the code. removed the hidden dynamic hidden field.

src/app/organisation/index.njk
SaritaMore7: I have applied 'govuk-button' style for <a> tag and added new style in the custom css file.

src/app/organisation/index.njk
SaritaMore7: It was loading endlessly  because of '<button>' tag. so changed it to <a>.

ame-test: Oh, cool,  out of curiousity, why is that the case?

ame-test (approved):
Approve, comments I did have been done.

src/app/organisation/index.njk
```diff
@@ -56,25 +58,21 @@
...
-      <div class="govuk-grid-column-full">
-        <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--xl govuk-!-margin-top-0">
-        <p class="govuk-body-m">Search for people within the organisation,  {{ cookie.getOrganisationName() }}</p>
-        <input class="govuk-input govuk-input--width-20" id="organisation-search" name="organisation-search" type="text" placeholder="Search" style="margin-bottom: 10px">
-      </div>
       <form id="organisation" action="/organisation" class="form-group" method="post" role="form" >
+        <div class="govuk-grid-column-two-thirds">    
+          <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--xl govuk-!-margin-top-0">
+          <p class="govuk-body-m">Search for people within the organisation,  {{ cookie.getOrganisationName() }}</p>
+          <input class="govuk-input govuk-input--width-20" id="organisation-search" name="organisation_search" type="text" value="{{ searchUserName }}" placeholder="Search">
```
ame-test, line 65: Bug: If you go and search for a user and then (with the value in the search field) decide to edit or delete a value, it will return straight back to the search page:
* reproduce:
  * Search a value.
  * Click on edit ro delete for a particular user.
  * Return to search page not edit ro delete (see screenshots).
  
  
  Instead, you need handle the routing of the `organisation_search` property in the post.controller.js differently.
  * in other parts of the app, `formaction` [attribtues](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#formaction) have been used, but if there a better solution, please use it.

ame-test (changes requested):
Bug comment to look into.

src/app/organisation/index.njk
ame-test: Screenshot 2024-08-07 at 12 47 47
Screenshot 2024-08-07 at 12 47 53
Screenshot 2024-08-07 at 12 48 02
Screenshot 2024-08-07 at 12 48 18

src/test/organisation/post.controller.test.js
```diff
@@ -27,13 +29,35 @@ describe('Organisation Post Controller', () => {
       redirect: sinon.spy(),
     };
 
+    paginationStub = sinon.stub(pagination, 'setCurrentPage');
     saveSessionStub = sinon.stub(req.session, 'save').callsArg(0);
   });
 
   afterEach(() => {
     sinon.restore();
   });
```
ame-test, line 39: If one was continuing with send the `organisation_search` stuff to the `post.controler.js` it should be covered here.

src/test/organisation/searchusers/get.controller.test.js
```diff
@@ -0,0 +1,104 @@
...
+        { id: 'PERSON-4', first_name: 'PERSON-1-name', role: { name: 'User' } },
+        { id: 'PERSON-5', first_name: 'PERSON-3-name', role: { name: 'Admin' } },
+      ],
+    }));
+    const callController = async () => {
+      await controller(req, res);
+    };
+
+    callController().then(() => {
+      expect(res.render).to.not.have.been.calledOnceWithExactly('/organisation/index', {
```
ame-test, line 93: Technically speaking:
* `errMsg` and `successMsg` route should be here.

src/test/organisation/post.controller.test.js
SaritaMore7: done. added new test case for organisation_search.

src/test/organisation/searchusers/get.controller.test.js
```diff
@@ -56,6 +57,72 @@ describe('Organisation Search Users Get Controller', () => {
...
+      expect(req.session.successHeader).to.be.undefined;
+      expect(orgApiStub).to.have.been.calledOnceWithExactly('ORGANISATION-ID-1','TEST');
+      expect(res.render).to.not.have.been.calledOnceWithExactly('/organisation/index', {
+        cookie,
+        orgUser: [
+          { id: 'PERSON-1', first_name: 'PERSON-1-name', role: { name: 'User' } },
+          { id: 'PERSON-4', first_name: 'PERSON-1-name', role: { name: 'User' } },
+          { id: 'PERSON-5', first_name: 'PERSON-3-name', role: { name: 'Admin' } },
+        ],
+        successHeader: 'Successful header',
```
ame-test, line 119: Question: Where is the `successHeader` defined in the this test?
* I can't find in the file, but I could be misunderstand thing.

ame-test (approved):
This is good, prove of edit user would be great, toher wise comment on unit test, but I don't consider it blocking.

src/app/organisation/index.njk
SaritaMore7: image


**Edit page**

image


**After Edit**

image

PR #526, opened 2024-08-12
FAIR-614 (Responsible person table data is displayed under the wrong …

sjose012:
…table header)

**Definition of Done**

### DIFF
--- a/src/app/resperson/index.njk
+++ b/src/app/resperson/index.njk
@@ -56,10 +56,10 @@
                       <button id="edit-edit_reponsiblePerson-{{ loop.index }}" class="govuk-link btn-link" 
                         title="Edit details for {{ resPerson.responsibleGivenName }} {{ resPerson.responsibleSurname }}" 
                         formaction="resperson/edit" formmethod="get"
-                        name="editResponsiblePerson" value="{{resPerson.responsiblePersonId}}">{{resPerson.responsibleGivenName}}
+                        name="editResponsiblePerson" value="{{resPerson.responsiblePersonId}}">{{resPerson.responsibleSurname}}
                       </button>
                     </td>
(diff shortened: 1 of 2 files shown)


src/app/resperson/index.njk
```diff
@@ -56,10 +56,10 @@
                       <button id="edit-edit_reponsiblePerson-{{ loop.index }}" class="govuk-link btn-link" 
                         title="Edit details for {{ resPerson.responsibleGivenName }} {{ resPerson.responsibleSurname }}" 
                         formaction="resperson/edit" formmethod="get"
-                        name="editResponsiblePerson" value="{{resPerson.responsiblePersonId}}">{{resPerson.responsibleGivenName}}
+                        name="editResponsiblePerson" value="{{resPerson.responsiblePersonId}}">{{resPerson.responsibleSurname}}
                       </button>
                     </td>
-                    <td class="govuk-table__cell" scope="row">{{resPerson.responsibleSurname}}</td>
+                    <td class="govuk-table__cell" scope="row">{{resPerson.responsibleGivenName}}</td>
```
ame-test, line 62: Nit: `scope="row"` should be on the surname element than the resperson given name if it is to be used.

ame-test, line 62: Nit: `scope="row" is not for `td` elements, is is for `th` elements.
* It is deprecated (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/td#scope).

ame-test (approved):
Approved, nit comment is nit.

src/app/resperson/index.njk
sjose012: removed the scope from td element

ame-test: Thanks.

PR #557, opened 2025-02-24
NMSW-2691-revised design of passenger check page

SaritaMore7:
Added new revised design for 'Passenger check' page.

image

**Definition of Done**


src/app/garfile/amg/checkin/pane.njk
```diff
@@ -48,6 +49,9 @@
           {% else %}
             There was an error processing the permission to travel for this person.
           {% endif %}
+          {% else %}
+           <span style="opacity:.5;"> Awaiting ...</span>
```
ame-test, line 53: `nit` `accessibility`: This colour seems like it will fail accessibility AA colour contrast.
* Is this ok or not?

I get not fixing it.

If fixing it, also fix below and any other ones.:
```html
<td class="govuk-table__cell" colspan="2" style="opacity:.5;">Awaiting Permission to Travel responses...</td>
```

src/app/garfile/amg/checkin/pane.njk
```diff
@@ -80,28 +84,22 @@
...
+<div class="margin-bottom-40 govuk-body">
+  <p>To evidence a passenger's immigration status, you must:</p>
+  <ul>
+  <li>check if they are <a href="https://www.gov.uk/government/publications/uk-visa-requirements-list-for-carriers/uk-visa-requirements-for-international-carriers">from a country that needs a visa to travel to the UK</a></li>
+  <li>review their physical documentation or <a href="https://www.gov.uk/check-immigration-status">use a sharecode</a> they have provided in advance</li>
+  </ul>
+   
+   <p>If you cannot get valid digital or physical evidence from the passenger, you can contact Border Force for advice. The Carrier Support Hub (CSH) is open 24/7.</p>
+   <ul>
+      <li>UK: 0300 369 0610</li>
```
ame-test, line 100: `nit`: Please use the `CARRIER_SUPPORT_HUB_UK_NUMBER` and the `CARRIER_SUPPORT_HUB_INTERNATIONAL_NUMBER` environment variables for the carrier code phone numbers.
* Allows for phone number to be changed without code change.
* Also, means in code based, does not show the phone number (remember the whole phone number incident).

src/app/garfile/amg/checkin/pane.njk
SaritaMore7: This message is to inform the user that the page is currently loading. It will disappear once the AMG statuses have finished loading.

src/app/garfile/amg/checkin/pane.njk
SaritaMore7: done

PR #564, opened 2025-03-16
Nmsw 2830

ame-test:
- `package-lock.json` used for reasons --> move to multi-stage build.
- major node.js bump to node 22 and alpine increase, removes `cross-spawn` issue.
- `nanoid` upgrade, uses `utils` implementation.
- removed unused dependency `staticify`, not properly implemented to do any of the static stuff.

2 issues left, `csurf` and `requests`, both harder issues to fix.

**Definition of Done**


SaritaMore7:
The upgrade changes look good to me. We should perform regression testing after merging the code into the main branch. 

The two packages 'csurf' and 'requests' which have been deprecated for the past five years and  should be handled in a separate ticket. Upgrading them will require significant changes. we need to find alternatives for them.

SaritaMore7 (approved):
approved!

PR #575, opened 2025-05-19
NMSW-3120-sgar-onelogin-logout

ndifrekeekottHO:

**Definition of Done**


src/app/user/onelogin/post.controller.js
```diff
@@ -72,7 +72,7 @@ async function handleGivenNameSubmission(req, res) {
       errors = validationError
       logger.error("Validation error");
 
-      return [Outcome.VALIDATION_FAILED, {firstName, lastName, errors}, null];
+      return [Outcome.VALIDATION_FAILED, {firstName: firstName, lastName, errors}, null];
```
SaritaMore7, line 75: could you please replace {**firstName: firstName,** } with {firstName}

ndifrekeekottHO, line 75:
```suggestion
      return [Outcome.VALIDATION_FAILED, {firstName, lastName, errors}, null];
```

SaritaMore7 (approved):
tested logout functionality on localhost and it works as expected.

PR #588, opened 2025-06-09
feat: Add validation rules to form

ndifrekeekottHO:

**Definition of Done**

Screenshot 2025-06-09 at 14 51 12
Screenshot 2025-06-09 at 14 50 58
Screenshot 2025-06-09 at 14 50 43

### DIFF
--- a/src/common/utils/validator.js
+++ b/src/common/utils/validator.js
@@ -36,7 +36,7 @@ function isOtherDocumentWithDocumentDesc(args){
   if (!args) return false;
   const [documentType, documentDesc] = args;
 
-  return isEmpty(documentDesc) || documentType === "Other" 
+  return isEmpty(documentDesc) || documentType === "Other"
 }
 
 /**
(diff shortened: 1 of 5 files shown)


src/common/utils/validator.js
```diff
@@ -125,6 +125,10 @@ function validName(value) {
   return regex.test(value);
 }
 
+function nameHasNoNumbers(value) {
+  return !/\d/.test(value);
+}
+
```
SaritaMore7, line 131: why don't you  use 'validName' function to validate the name fields?

ndifrekeekottHO: Because we have a specific Error message in the mockup that flags for Numbers. `validName` handles both special character and numbers. There is a different message for that.

ndifrekeekottHO: Screenshot 2025-06-09 at 16 18 34
Screenshot 2025-06-09 at 16 18 18

PR #593, opened 2025-06-19
nmsw-2949 added UI pagination to dashboard

SamTaylorHO:
https://jira.bics-collaboration.homeoffice.gov.uk/browse/NMSW-2949

Added two types of pagination to the GAR dashboard. 
The first is client side pagination for the first 10,000 results, currently all 10,000 are shown on the screen at once. This change allows the user to paginate at with 10 results per page. Server side pagination was not possible as there were many defects with the tab widgets behaviour when both the tab widget and the paginator are using the URL to track state.

Screenshot 2025-06-20 at 09 19 19

Screenshot 2025-06-20 at 09 19 08

Screenshot 2025-06-20 at 09 19 13

The second pagination is client side pagination, this allows us to remove the confusing "you have x remaining GARs" message. In the incredibly unlikely scenario that the user has more than 10,000 GARs (we wipe GARs after 6 months), an additional pagination option is show that will allow the user to load the next 10,000 results from the server. In this case the defects with the tab widget are less noticeable, as the pagination will be less frequent. It is also highly unlikely a user will ever see this behaviour.

Screenshot 2025-06-19 at 16 22 11

**Definition of Done**


src/app/home/index.njk
```diff
@@ -56,8 +58,20 @@
...
         <div class="govuk-inset-text">Your draft, submitted and cancelled GARs will only be saved for 6 months, after this they will be deleted.</div>
+        {% if serverPagination.totalPages>1 %}
+        <div class="govuk-inset-text">
+          You are viewing {{serverPagination.perPage * (serverPagination.page-1) +1}} - {{serverPagination.totalItems if serverPagination.page == serverPagination.totalPages else serverPagination.perPage * serverPagination.page}}  of your most recent GARs, there are {{serverPagination.totalItems}} total GARs. 
+          {% if serverPagination.page>1 %}
+          <br/>To view the previous {{serverPagination.perPage}} GARs <a href="?page={{serverPagination.page-1}}">click here</a>.
+          {% endif %}
+          {% if serverPagination.page<serverPagination.totalPages %}
+          <br/>To view the next {{serverPagination.totalItems % serverPagination.perPage if serverPagination.page == serverPagination.totalPages-1 else serverPagination.perPage}} GARs <a href="?page={{serverPagination.page+1}}">click here</a>.
+          {% endif %}
```
ndifrekeekottHO, line 69: :suggestion: 

I think we can put certain computations to variables to make them clearer. Example: 

`{% set hasMorePages = serverPagination.page < serverPagination.totalPages %}`

This can then be used in the line: 

```
{% if hasMorePages %}
<br/>To view the next {{ nextPages  if lastPage else serverPagination.perPage}} GARs <a href="?page={{serverPagination.page+1}}">click here</a>.
{% endif %}
```

Main highlight, there is a lot of computation mixed within, setting variables at the top will simplify the tags. Let me know what you think.

PR #595, opened 2025-06-20
NMSW-3095 - fixed the 0B content alignment issue

SaritaMore7:

image

**Definition of Done**

### DIFF
--- a/src/public/stylesheets/custom.css
+++ b/src/public/stylesheets/custom.css
@@ -202,6 +202,6 @@ h2.govuk-notification-banner__title {
     height: 6.5px;
 }
 
-.col-vertical-alignment{
+.vertical-align-top {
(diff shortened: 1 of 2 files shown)


src/public/stylesheets/custom.css
```diff
@@ -201,3 +201,7 @@ h2.govuk-notification-banner__title {
     width: 10px;
     height: 6.5px;
 }
+
+.col-vertical-alignment{
```
ho-henry-senior, line 205: could this be called .vertical-align-top ? 
- it applies to cells rather than columns
-  putting 'top' in the name shows the effect it will have?

SaritaMore7: done

ho-henry-senior (approved):

PR #609, opened 2025-07-08
NMSW-3165--update-admin-notify-email

ndifrekeekottHO:

**Definition of Done**


src/app/user/login/get.controller.js
```diff
@@ -186,6 +198,8 @@ module.exports = (req, res) => {
             return Promise.reject();
           }
 
+          accountUrl = parseUrlForNonProd(req, accountUrl);
```
ndifrekeekottHO, line 201: @SaritaMore7  updating url here.

SaritaMore7 (approved):
approved

PR #630, opened 2025-07-25
feat: Add returned data after update

ndifrekeekottHO:

return updated user for redirect

**Definition of Done**


src/app/user/login/get.controller.js
```diff
@@ -109,13 +109,13 @@ const handleUserAuthentication = (userInfo, cookie) => {
           );
         case emailMatches && userData.oneLoginSid === null:
           // user email exists, but onelogin is null or does not match - action, update SID
-          return new Promise((resolve, reject) => {
+          return new Promise((resolve, reject) =>
```
ndifrekeekottHO, line 112: not sure how but braces were added and data no returned.

PR #636, opened 2025-07-29
NMSQ-3355-prevent-double-submission-create-account-btn

ndifrekeekottHO:
- prevent double submission. 

**Definition of Done**

### DIFF
--- a/src/common/templates/base-template.njk
+++ b/src/common/templates/base-template.njk
@@ -209,7 +209,6 @@
           .appendTo("#page-form");
         return true;
       });
-    });
     </script>
     <script src="/javascripts/table-multiselect.js"></script>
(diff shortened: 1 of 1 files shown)


src/common/templates/base-template.njk
```diff
@@ -209,6 +209,25 @@
...
+            form.addEventListener("submit", (e) => {
+                // Check if the form has already been submitted
+                if (isFormSubmitted) {
+                    e.preventDefault(); // Prevent the form from submitting again
+                } else {
+                    isFormSubmitted = true; // Mark the form as submitted
+                }
+            });
+        });
+    });
```
SaritaMore7, line 230: can you put this code into separate javascript file under public/javascript/preventDoubleClick.js and then include the script path here

PR #653, opened 2025-09-30
Nmsw 3464 email configuration for 0T

SaritaMore7:
Added link in the email to access the GAR summary. When user clicks on the email link, they are taken directly to the summary page.

**Definition of Done**


src/common/templates/check-your-answers.njk
```diff
@@ -1,5 +1,5 @@
 {% import "common/templates/includes/banners.njk" as b %}
-
+{% set viewGuidance = (garfile.status.name === 'Submitted' and resubmitted != 'yes' and numberOf0TResponseCodes > 0 and durationInDeparture > 125) %}
```
ndifrekeekottHO, line 2: ❓  Any reason `resubmitted` is set to the values `no|yes` rather than `true|false`?

PR #810, opened 2026-04-09
(INC6010928) NMSW 4149 GAR file upload

TemitopeAyokuHO:
[NMSW-4149](https://jira.bics-collaboration.homeoffice.gov.uk/browse/NMSW-4149) - GAR file upload issue (INC6010928) - add validation to remove the whitespaces.

**Definition of Done**

### DIFF
--- a/src/common/utils/validator.js
+++ b/src/common/utils/validator.js
@@ -430,9 +430,29 @@ function realDateInFuture(dObj) {
   );
 }
 
+const dateTransformer = (str) => {
+  if (typeof str !== 'string') {
+    return str;
+  }
+
+  const match = /^(\d{2})\/(\d{2})\/(\d{4})$/.test(str);
+
+  if (match) {
+    const [day, month, year] = str.split('/');
+
+    if (validDay(day, month, year) && validMonth(month) && validYear(year)) {
+      return `${year}-${month}-${day}`;
+    }
+  }
+  return str;
+};
+
 function realDateFromString(str) {
   moment.suppressDeprecationWarnings = true;
-  return moment(str).isValid();
+  if (str instanceof Date) {
+    return moment(str).isValid();
+  }
+  return moment(dateTransformer(str)).isValid();
 }
 
 // This function will validate Passport Expiry date while uploading Gar Template
--- a/src/test/api/uploadgar/post.controller.test.js
+++ b/src/test/api/uploadgar/post.controller.test.js
@@ -199,7 +199,7 @@ describe('API upload GAR post controller', () => {
         await controller(req, res);
       };
 
-      callController().then(() => {
+      await callController().then(() => {
         expect(req.session.save).to.have.been.called;
         expect(req.session.failureMsg).to.eql([
           new ValidationRule(
@@ -270,7 +270,7 @@ describe('API upload GAR post controller', () => {
         await controller(req, res);
       };
 
-      callController().then(() => {
+      await callController().then(() => {
         expect(req.session.save).to.have.been.called;
         expect(req.session.failureMsg).to.eql([
           new ValidationRule(validator.isValidAirportCode, '', null, 'Arrival port should be an ICAO or IATA code'),
@@ -290,7 +290,7 @@ describe('API upload GAR post controller', () => {
         await controller(req, res);
       };
 
-      callController().then(() => {
+      await callController().then(() => {
         expect(req.session.save).to.have.been.called;
         expect(req.session.failureMsg).to.eql([
           new ValidationRule(
@@ -329,7 +329,7 @@ describe('API upload GAR post controller', () => {
         });
     });
 
-    it('should return with an error when api returns a message', () => {
+    it('should return with an error when api returns a message', async () => {
       const data = getValidWorkbook();
 
       sinon.stub(createGarApi, 'createGar').resolves(
@@ -345,7 +345,7 @@ describe('API upload GAR post controller', () => {
         await controller(req, res);
       };
 
-      callController().then(() => {
+      await callController().then(() => {
         expect(createGarApi.createGar).to.have.been.calledWith('khan@augmented.com');
         expect(garApi.patch).to.not.have.been.called;
         expect(req.session.save).to.have.been.called;
@@ -355,7 +355,7 @@ describe('API upload GAR post controller', () => {
 
     // TODO: Error message informs the GAR could not be updated, but one was created anyways?
     // TODO: There are three calls to update the GAR, but payload could be combined into one? Maybe?
-    it('should mention when on of the update steps rejects', () => {
+    it('should mention when on of the update steps rejects', async () => {
       const data = getValidWorkbook();
 
       sinon.stub(createGarApi, 'createGar').resolves(
@@ -375,7 +375,7 @@ describe('API upload GAR post controller', () => {
       };
 
       // Promise.all for three Promises...
-      callController()
+      await callController()
         .then()
         .then()
         .then()
@@ -423,7 +423,7 @@ describe('API upload GAR post controller', () => {
         });
     });
 
-    it('Phoney document type shoudl be disallowed', () => {
+    it('Phoney document type shoudl be disallowed', async () => {
       const data = getValidWorkbook();
 
       data.Sheets.Valid1.A9.v = 'Cipher';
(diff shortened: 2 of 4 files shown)


src/common/utils/validator.js
```diff
@@ -431,8 +431,12 @@ function realDateInFuture(dObj) {
 }
 
 function realDateFromString(str) {
+  if (typeof str !== 'string') {
+    return str;
+  }
+  const match = str.split(/^(\d{1,2}\/\d{1,2}\/\d{4})/).filter(Boolean);
   moment.suppressDeprecationWarnings = true;
-  return moment(str).isValid();
+  return moment(match).isValid();
```
ndifrekeekottHO: @TemitopeAyokuHO 

Though I think this code works. However, we did have a previous discussion about moment. The project is no longer in active development. Can we instead use inbuilt Date functions/methods to achieve the same result? 

Is it possible to do that with native functionalities? 

cc: @jamiecarterHO  @SaritaMore7

ndifrekeekottHO (changes requested):

src/common/utils/validator.js
```diff
@@ -431,8 +431,12 @@ function realDateInFuture(dObj) {
 }
 
 function realDateFromString(str) {
+  if (typeof str !== 'string') {
+    return str;
+  }
```
jamiecarterHO, line 436: Are we expecting a bool response from this function? If so should probably `return false` here

src/common/utils/validator.js
jamiecarterHO, line 1: This change will validate dates with extra spaces but it puts the onus on the user to understand what the issue is from our error message and correct it themselves in the template. This may not be easy to do. 

It may be more user friendly to trim dates before validation and storage and handle issues automatically.

src/common/utils/validator.js
TemitopeAyokuHO: @jamiecarterHO the function is returning date

jamiecarterHO: I believe `moment(value).isValid()` will return `true` or `false`

https://github.com/UKHomeOffice/egar-public-site-ui/blob/dev/src/test/validators.test.js#L562
https://momentjs.com/docs/#/parsing/string/

src/common/utils/validator.js
```diff
@@ -431,8 +431,17 @@ function realDateInFuture(dObj) {
 }
 
 function realDateFromString(str) {
-  moment.suppressDeprecationWarnings = true;
-  return moment(str).isValid();
+  if (typeof str !== 'string') {
+    return false;
+  }
+  const trimWhiteSpace = str.trim();
+  const match = trimWhiteSpace.split(/^(\d{1,2}\/\d{1,2}\/\d{4})/).filter(Boolean);
```
jamiecarterHO: This will still trim and filter a date only for validation, but the date string will still be stored with the additional whitespace.

Think we need to transform the date string before it comes to this validation

TemitopeAyokuHO: I have refactored this in different function

src/common/utils/validator.js
```diff
@@ -430,9 +430,27 @@ function realDateInFuture(dObj) {
...
+  }
+  return str;
+};
+
 function realDateFromString(str) {
+  const dateString = dateTransformer(str);
+
+  if ((!dateString) instanceof Date) {
+    return false;
+  }
```
jamiecarterHO: This will check dateString's truthiness first then check if that bool is an instance of Date...not sure if this block can ever be triggered.

Would it work if we do?

```javascript
moment.suppressDeprecationWarnings = true;
if (!(str instanceof Date)) {
  return moment(dateTransformer(str)).isValid();
}
return moment(str).isValid();

File: src/common/utils/validator.js [hunk 5]
```diff
@@ -430,9 +430,27 @@ function realDateInFuture(dObj) {
   );
 }
 
+const dateTransformer = (str) => {
+  if (typeof str !== 'string') {
+    return str;
+  }
+  const [day, month, year] = str.split('/');
```
jamiecarterHO: Perhaps it would be good to check that `str` is in an acceptable format to begin with? Something like

`const match = /^(\d{2})\/(\d{2})\/(\d{4})$/.test(str);`

I think maybe if we split a date like `'23-02-2026'` then `month` and `year` are undefined which may have unexpected consequences

src/test/api/uploadgar/post.controller.test.js
```diff
@@ -303,12 +363,62 @@ describe('API upload GAR post controller', () => {
         expect(res.redirect).to.have.been.calledWith('/garfile/garupload');
       });
     });
+
+    it('return message if date is invalid with space', async () => {
```
jamiecarterHO: This is a fine test for when there is a standard space but the issue we are looking at would involve dates that were inputted like `'12/12/2026\u00A0'` (format can also be read as mm/dd/yyyy and finish on a no-break space). In those cases there should be no error and the `dateTransformer` should turn the badly formatted date into one that can be accepted.

src/test/api/uploadgar/post.controller.test.js
```diff
@@ -187,6 +187,30 @@ describe('API upload GAR post controller', () => {
       expect(req.session.save).to.have.been.called;
       expect(req.session.failureMsg).to.eql([
         new ValidationRule(validator.validGender, '', 'Gender', 'Enter a valid sex for crew member James Kirk'),
+        new ValidationRule(
```
jamiecarterHO: I'm not entirely sure why we need these extra `ValidationRule` where new updates aren't being tested - did these start failing?

src/test/api/uploadgar/post.controller.test.js
```diff
@@ -241,6 +289,10 @@ describe('API upload GAR post controller', () => {
 
       const data = getValidWorkbook();
       data.Sheets.Valid1.D3.v = '2022-07-30';
+      data.Sheets.Valid1.H20.v = '10/10/1965';
+      data.Sheets.Valid1.K20.v = '10/13/2033';
+      data.Sheets.Valid1.H9.v = '10/10/1965';
+      data.Sheets.Valid1.K9.v = '10/13/2033';
```
jamiecarterHO: Not sure why we need to add these extra cells to the test doc setup.

TemitopeAyokuHO: That will trigger the test to fail since the validations are enforced on those cell

src/test/api/uploadgar/post.controller.test.js
TemitopeAyokuHO: do you suggest to remove this test

src/test/api/uploadgar/post.controller.test.js
TemitopeAyokuHO: yes as mentioned if the date field is empty or not in correct format,, it will fail test

src/common/utils/validator.js
```diff
@@ -430,9 +430,29 @@ function realDateInFuture(dObj) {
...
+      return `${year}-${month}-${day}`;
+    }
+  }
+  return str;
+};
+
 function realDateFromString(str) {
   moment.suppressDeprecationWarnings = true;
-  return moment(str).isValid();
+  if (!(str instanceof Date)) {
```
jamiecarterHO: If argument `str` is an instance of `Date` then we want to check `str` against moment without transformation - I think then you need to remove the exclamation mark here.

src/test/api/uploadgar/post.controller.test.js
jamiecarterHO: Is that not because the date in `data.Sheets.Valid1.D3.v` (or any date got from `getValidWorkbook()` in the format yyyy-mm-dd) now does not pass validation? Maybe they should be changed in the source test data rather than test by test?

src/common/utils/validator.js
TemitopeAyokuHO: done

src/test/api/uploadgar/post.controller.test.js
jamiecarterHO: You could avoid altering all the tests by updating the dates in the cells updated here in src/test/api/uploadgar/workbook-data.js

jamiecarterHO (approved):
Looks good to me!

ndifrekeekottHO (approved):

PR #812, opened 2026-04-13
Added an official ECR plugin to drone.yml and removed existing steps …

b-pb:
…of build & push

**Definition of Done**

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -63,7 +63,7 @@ steps:
       IMAGE_NAME: egar-public-site-ui:${DRONE_COMMIT}
       IGNORE_UNFIXED: true
     depends_on:
-    - publish-to-ecr
+    - build
 
   - name: sonar-scan
     pull: if-not-exists
--- a/.drone.yml
+++ b/.drone.yml
@@ -8,24 +8,11 @@ platform:
   arch: amd64
 
 steps:
-  - name: publish-to-ecr
-    image: plugins/ecr
-    settings:
-      access_key:
-        from_secret: DEV_EGAR_PUBLIC_SITE_UI_AWS_ACCESS_KEY_ID
-      secret_key:
-        from_secret: DEV_EGAR_PUBLIC_SITE_UI_AWS_SECRET_ACCESS_KEY
-      region: eu-west-2
-      repo: egar/public-site-ui
-      registry: 340268328991.dkr.ecr.eu-west-2.amazonaws.com
-      tags:
-        - ${DRONE_COMMIT_SHA}
-    when:
-      event:
-        - push
-      branch:
-        - main
-        - dev
+  - name: build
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -1,36 +1,25 @@
...
 platform:
   os: linux
   arch: amd64
 
 steps:
-  - name: build
-    pull: if-not-exists
-    image: 340268328991.dkr.ecr.eu-west-2.amazonaws.com/acp/dind:latest
-    commands:
-      - source ./ops/image_build.sh
```
jamiecarterHO, line 18: We'll still need the build step as we want to build on new PR trigger for scans before a push to ECR.

.drone.yml
```diff
@@ -68,7 +57,7 @@ steps:
       IMAGE_NAME: egar-public-site-ui:${DRONE_COMMIT}
       IGNORE_UNFIXED: true
     depends_on:
-    - build
+    - publish-to-ecr
```
jamiecarterHO: We want to scan on build still.

jamiecarterHO (changes requested):

.drone.yml
```diff
@@ -56,6 +32,25 @@ steps:
...
+    settings:
+      access_key:
+        from_secret: DEV_EGAR_PUBLIC_SITE_UI_AWS_ACCESS_KEY_ID
+      secret_key:
+        from_secret: DEV_EGAR_PUBLIC_SITE_UI_AWS_SECRET_ACCESS_KEY
+      region: eu-west-2
+      repo: egar/public-site-ui
+      registry: 340268328991.dkr.ecr.eu-west-2.amazonaws.com
+      tags:
+        - ${DRONE_COMMIT_SHA}
```
jamiecarterHO: I think in this case we should use the variable `DRONE_COMMIT` as this is used elsewhere in scripts. They should have the same value, but I think we should be consistent.

.drone.yml
```diff
@@ -56,6 +32,25 @@ steps:
     - npm install
     - npm run check
 
+  - name: publish-to-ecr
```
jamiecarterHO, line 35: we want this step to:

```
depends_on:
  - build
```

jamiecarterHO (changes requested):

.drone.yml
b-pb: Yes indeed, good catch Jamie.

.drone.yml
jamiecarterHO: Actually I think this is not needed us the ecr plugin has a build step and nothing else is dependent. What do you think @b-pb ?

b-pb (commented):
Changed depends_on:
    - build

jamiecarterHO (approved):
Ok lets try it!

PR #819, opened 2026-04-28
NMSW 3966 Account screen mobile responsiveness and banner alignment

TemitopeAyokuHO:
[NMSW-3966](https://jira.bics-collaboration.homeoffice.gov.uk/browse/NMSW-3966) - Account screen mobile responsiveness and banner alignment

- Added media query to help visualise the app in responsive mode 

Screenshot 2026-04-28 at 13 37 10

Screenshot 2026-04-28 at 13 37 31

**Definition of Done**

### DIFF
--- a/src/public/stylesheets/custom.css
+++ b/src/public/stylesheets/custom.css
@@ -1,46 +1,57 @@
-.lat-long:invalid{
-    background-color:#ff9baf;
+.lat-long:invalid {
+    background-color: #ff9baf;
 }
+
 .autocomplete__wrapper {
-  font-family: "GDS Transport", Arial, sans-serif;
+    font-family: "GDS Transport", Arial, sans-serif;
 }
+
 ::backdrop {
     background: black;
     opacity: 0.5;
 }
+
 .warning-button {
     background: #b10e1e !important;
     color: white !important;
 }
+
 .table-clickable tbody .unselectable-row:hover {
     background-color: inherit;
     cursor: default;
 }
+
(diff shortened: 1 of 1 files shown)


src/public/stylesheets/custom.css
```diff
@@ -192,57 +211,71 @@ h2.govuk-notification-banner__title {
...
     font-size: 19px !important;
 }
 
+
+@media (min-width: 300px) {
+    .govuk-grid-row {
+        margin-right: 30px;
+        margin-left: -5px;
+    }
+}
```
jamiecarterHO: the `govuk-grid-row` class should be responsive already at any width so my suspicion is that something else is causing the margin offset reported. 

Is there some some nested html issue? Think that may be a better solution than blanket forcing all rows to behave like this.

jamiecarterHO (changes requested):
We will be merging the existing GOV.UK frontend v5 branch into dev soon, and so this PR will have the correct target then, but for now this change is being made against code that may already have been updated.

Probably best to change the target to the govuk5-test branch and see if anything changes, or wait until after the merge into dev.

src/public/stylesheets/custom.css
TemitopeAyokuHO: Yea you right , after checking the whole css file i realised most of the style sheet has been modified which leave me to no option to create a query media for .govuk-grid-row .

TemitopeAyokuHO:
This will be closed as the changes. came through the gov-v5 change and ucd confirm layout is responsive

PR #825, opened 2026-05-05
feat: Reverting country selection to autocomplete rather than select

ndifrekeekottHO:

New form does not pre-select 
Screenshot 2026-05-05 at 12 31 53

Edit mode selects values

Screenshot 2026-05-05 at 12 32 05

Error screens 
Screenshot 2026-05-05 at 12 45 08
Screenshot 2026-05-05 at 12 45 03

Reverting the select back to auto-complete and stop pre-selecting first value

**Definition of Done**


src/common/templates/common/macros.njk
jamiecarterHO, line 1: This actually feels like a slightly better location, but there is already a generic macros file at src/common/templates/includes/macros.njk. Would this fit there?

jamiecarterHO (approved):

PR #835, opened 2026-05-18
NMSW-4783-Scan-for-egar-public-site-ui

ndifrekeekottHO:

Trivy current reports false positives  for these packages.  Example Picomatch, has been updated but it still gets flagged. 
Picomatch - https://security.snyk.io/package/npm/picomatch/2.3.2
brace-expansion - https://security.snyk.io/package/npm/brace-expansion

Screenshot 2026-05-18 at 17 06 12

**Definition of Done**


jamiecarterHO (approved):

Dockerfile
```diff
@@ -4,7 +4,7 @@ RUN apk update && apk upgrade --update-cache --available && apk add curl
 
 RUN mkdir -p /public-site/ && \
     mkdir -p /var/log/nodejs/ && \
-    touch /var/log/nodejs/app.log
+    touch /var/log/nodejs/app.log &&  npm install -g npm@11.15.0
```
jamiecarterHO: I think if we're trying to get past the vulns that come with the Node image's NPM version it might be better to use a trivy whitelist than force an NPM version or do it with npm overrides.

PR #838, opened 2026-05-19
nmsw-3795-sgar-fe-header-update

ndifrekeekottHO:

Had issues with the Mobile view of the page. The OneLogin Login menu was not showing.  However, the colour changes per browser: Yellow on Chrome, blue on Safari. 
Screenshot 2026-05-19 at 16 52 21

Shows Yellow background in chrome based browsers. 

Screenshot 2026-05-20 at 09 07 57

**Definition of Done**


src/public/javascripts/init-service-header.js
```diff
@@ -0,0 +1,91 @@
+/**
```
ndifrekeekottHO, line 1: This isn't hand written code. This was done following instructions from the repo

https://github.com/govuk-one-login/service-header/tree/v5.0.0#using-the-headers-javascript-file

jamiecarterHO (approved):

PR #842, opened 2026-05-21
feat: Change where Logo points to

ndifrekeekottHO:

Change Logo link redirect

**Definition of Done**

### DIFF
--- a/src/common/templates/includes/header-pre-user.njk
+++ b/src/common/templates/includes/header-pre-user.njk
@@ -1,7 +1,7 @@
 {% from "dist/govuk/components/header/macro.njk" import govukHeader %}
 
 {{ govukHeader({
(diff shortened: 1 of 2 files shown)


src/common/templates/includes/header-pre-user.njk
```diff
@@ -1,7 +1,7 @@
 {% from "dist/govuk/components/header/macro.njk" import govukHeader %}
 
 {{ govukHeader({
-  homepageUrl: "/"
+  homepageUrl: "http://www.gov.uk/"
```
jamiecarterHO:
```suggestion
  homepageUrl: "https://www.gov.uk/"
```

I'm sure it will upgrade the request but...

jamiecarterHO (approved):

PR #844, opened 2026-05-22
NMSW-4856 - updated dashboard tabs page code

SaritaMore7:
- updated tabs page code to fix the issue with pagination
- removed the paginator js code
- fixed mobile view issue
- fixed tabs gars table rows alignment 

image

image

**Definition of Done**


src/common/services/garApi.js
```diff
@@ -155,14 +155,17 @@ module.exports = {
    * @param {String} orgId id of organisation a user belongs to. Only required if org user.
    * @returns {Promise} Resolves with API response.
    */
-  getGars(userId, userType, page, orgId = null) {
+  getGars(userId, userType, pages, orgId = null) {
```
ho-henry-senior, line 158: could pages have a default? then you wouldn't need to declar a constant, and you wouldn't see `undefined` in your unit tests

src/app/home/get.controller.js
```diff
@@ -2,6 +2,7 @@ const logger = require('../../common/utils/logger')(__filename);
 const CookieModel = require('../../common/models/Cookie.class');
 const tokenApi = require('../../common/services/tokenApi');
 const garApi = require('../../common/services/garApi');
+const PER_PAGE = 10;
```
ho-henry-senior, line 5: I think this is ok as is, but there could be a more concise way to manage this default. a default in pageObj perhaps? or in garApi.getGars?

SaritaMore7: ok. will set this in the  garApi.getGars.

src/app/home/index.njk
```diff
@@ -29,92 +30,74 @@
...
+   
     <div class="govuk-grid-column-full">
+      
       <div class="govuk-tabs" data-module="govuk-tabs">
         <ul class="govuk-tabs__list">
-          <li class="govuk-tabs__list-item">
+          <li class="govuk-tabs__list-item status_tab">
             <a class="govuk-tabs__tab govuk-tabs__tab--selected" href="#gars-draft">
-              {{__('tab_gar_draft', draftGars.length)}}
+              {{__('tab_gar_draft', garsCountObj.Draft or 0 )}}
```
ho-henry-senior, line 52: could this defaulting to 0 happen in js code ?  again, just an opinion and happy with it as is

src/common/templates/includes/pagination.njk
```diff
@@ -116,15 +117,21 @@
...
+        {%if value and  key !== 'hash' %}
             {% set url = url ~ key ~ '=' ~ value ~ '&' %}
         {% endif %}
+        {% if key === 'hash' %}
+           {% set hash = value %}
+        {% endif %}
     {% endfor %}
 
-    {% set url = url ~ 'page=' ~ page %}
+    {% set url = url ~ 'page=' ~ page ~ hash %}
```
ho-henry-senior, line 131: does this mean the # would be passed into the `hash` value?

I think I would have put the # in the template and just passed the id, but maybe just a style thing; no need to change.

i.e. 
`{% set url = url ~ 'page=' ~ page ~ '#' ~ hash %}`

same as 
` {% set url = url ~ key ~ '=' ~ value ~ '&' %}` is doing?

src/public/javascripts/tabs-pagination.js
```diff
@@ -0,0 +1,10 @@
+const tabs = document.querySelectorAll('.status_tab');
```
ho-henry-senior: is this needed? did we discuss removing the module and simply using html links/href?

If possible, I think that would be better.

ho-henry-senior (commented):
some comments, let's chat later ?

src/common/templates/includes/pagination.njk
SaritaMore7: yes, it needs to be passed as an optional parameter if I need to append it to the end of the url. I can not achieve that by sending the id.

src/app/home/get.controller.js
```diff
@@ -14,33 +15,32 @@ module.exports = (req, res) => {
   // Delete any GAR stored in the cookie session
   cookie.session.gar = null;
 
+  const statusTab = req.query.status || 'Draft';
```
ndifrekeekottHO: suggestion: 

`const statusTab = req.query?.status || 'Draft'` 

Not sure if status is always present.

src/app/home/get.controller.js
```diff
@@ -14,33 +15,32 @@ module.exports = (req, res) => {
...
+
   tokenApi
     .getLastLogin(cookie.getUserEmail())
     .then((userSession) => {
       const { successHeader, successMsg } = req.session;
-      const page = req?.query?.page || 1;
+      const pageVal = req?.query?.page || 1;
```
ndifrekeekottHO: Remove magic numbers and assign to constant. 

:suggestion:

```javascript
// top of file
const PAGE_ONE = 1



const pageVal = req?.query?.page || PAGE_ONE
```

ndifrekeekottHO (changes requested):

ho-henry-senior (approved):

ndifrekeekottHO (approved):

PR #850, opened 2026-05-27
feat: Replacing request package

ndifrekeekottHO:

1. Replacing request packaging and making use of node's node-fetch utility instead. 

**Definition of Done**

### DIFF
--- a/src/app/garfile/view/get.controller.js
+++ b/src/app/garfile/view/get.controller.js
@@ -45,10 +45,12 @@ module.exports = async (req, res) => {
 
   try {
     cookie.setGarId(garId);
-    const garPeople = garApi.getPeople(garId);
-    const garDetails = dataAccessApi.garApi.get(garId);
-    const garDocs = garApi.getSupportingDocs(garId);
-    const { progress } = JSON.parse(await garApi.getGarCheckinProgress(garId));
+    const [garPeople, garDetails, garDocs, progress] = await Promise.all([
+      dataAccessApi.garApi.getPeople(garId),
+      dataAccessApi.garApi.get(garId),
+      dataAccessApi.garApi.getSupportingDocs(garId),
+      dataAccessApi.garApi.getGarCheckinProgress(garId),
+    ]);
 
     const resubmitted = req.query.resubmitted;
     const isResubmitted = cookie.getResubmitFor0T().includes(garId);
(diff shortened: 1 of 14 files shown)


src/app/garfile/view/get.controller.js
```diff
@@ -44,10 +45,12 @@ module.exports = async (req, res) => {
...
     cookie.setGarId(garId);
-    const garPeople = garApi.getPeople(garId);
-    const garDetails = garApi.get(garId);
-    const garDocs = garApi.getSupportingDocs(garId);
-    const { progress } = JSON.parse(await garApi.getGarCheckinProgress(garId));
+    const [garPeople, garDetails, garDocs, progress] = await Promise.all([
+      dataAccessApi.garApi.getPeople(garId),
+      dataAccessApi.garApi.get(garId),
+      dataAccessApi.garApi.getSupportingDocs(garId),
+      dataAccessApi.garApi.getGarCheckinProgress(garId),
```
jamiecarterHO, line 52: could destructure these down to the method or the service API?

src/app/garfile/view/get.controller.js
```diff
@@ -66,69 +69,65 @@ module.exports = async (req, res) => {
...
-        cookie.setGarId(parsedGar.garId);
-        cookie.setGarStatus(parsedGar.status.name);
-        logger.info(`Retrieved GAR id: ${parsedGar.garId}`);
+    try {
+      const parsedGar = garDetails;
+      const parsedPeople = garPeople;
+      const supportingDocuments = garDocs;
+      const { departureDate, departureTime } = parsedGar;
+      const lastDepartureDateString = departureDate && departureTime ? `${departureDate}T${departureTime}.000Z` : null;
+      const durationInDeparture = garApi.getDurationBeforeDeparture(departureDate, departureTime);
```
jamiecarterHO, line 78: Is this one now `dataAccessApi.garApi.getDurationBeforeDeparture`?

src/app/garfile/view/get.controller.js
```diff
@@ -66,69 +69,65 @@ module.exports = async (req, res) => {
...
-          resubmitted,
-          durationInDeparture,
-          numberOf0TResponseCodes,
-          isResubmitted,
-        };
-        renderContext.showChangeLinks = true;
-        if (parsedGar.status.name === 'Submitted' || parsedGar.status.name === 'Cancelled') {
-          renderContext.showChangeLinks = false;
-        }
+      // Maybe not necessary but delete the ids as the template does not need them
```
jamiecarterHO, line 93: Good to tidy up :)

src/common/services/dataAccessApi.js
```diff
@@ -0,0 +1,13 @@
...
+const { API_VERSION } = settings;
+const BASE_URL = `${API_BASE}${API_VERSION}`;
+const ApiClient = require('./httpClient');
+const { GarApi } = require('./garApi');
+
+const httpClient = new ApiClient({ baseUrl: BASE_URL });
+
+module.exports = {
+  garApi: new GarApi(httpClient),
+};
```
jamiecarterHO, line 13: Is the idea to extend this export to include all the service APIs with the others to be added later?

src/test/garfile/view/post.controller.test.js
```diff
@@ -130,10 +131,10 @@ describe('GAR view post controller', () => {
     callController()
       .then()
       .then(() => {
-        expect(garApiGetStub).to.have.been.calledOnceWithExactly('GAR-ID-EXAMPLE-1', true);
-        expect(garApiGetPeopleStub).to.have.been.calledOnceWithExactly('GAR-ID-EXAMPLE-1');
-        expect(garApiGetSupportingDocsStub).to.have.been.calledOnceWithExactly('GAR-ID-EXAMPLE-1');
-        expect(res.redirect).to.have.been.calledOnceWithExactly('/home');
+        // expect(garApiGetStub).to.have.been.calledOnceWithExactly('GAR-ID-EXAMPLE-1', true);
+        // expect(garApiGetPeopleStub).to.have.been.calledOnceWithExactly('GAR-ID-EXAMPLE-1');
+        // expect(garApiGetSupportingDocsStub).to.have.been.calledOnceWithExactly('GAR-ID-EXAMPLE-1');
```
jamiecarterHO: Can we remove them if they're not required?

src/common/services/garApi.js
```diff
@@ -4,17 +4,128 @@ const endpoints = require('../config/endpoints');
...
+    } catch (err) {
+      const responseErrorMessage = getResponseErrorMessage(err);
+      logger.error(`${garId} garApi.progress request was not successful : ${responseErrorMessage}`);
+      return responseErrorMessage;
+    }
+  }
 }
 
 module.exports = {
+  GarApi,
```
jamiecarterHO, line 129: `module.exports.GarApi.get()` uses `fetch` but `module.exports.get()` is still using `request` directly? Can we flatten those methods for `fetch` and call any `garApi` method from `dataAccessApi.garApi`?

src/common/services/httpClient.js
```diff
@@ -0,0 +1,89 @@
...
+    this.defaultHeaders = {
+      'Content-Type': 'application/json',
+      Accept: 'application/json',
+      ...defaultHeaders,
+    };
+
+    this.timeout = timeout;
+  }
+
+  async request(path, { method = 'GET', headers = {}, body, query } = {}) {
```
jamiecarterHO: If we're not using `request` perhaps we can save the noun `request` for express/node requests and call this method something else to differentiate?

src/common/services/httpClient.js
```diff
@@ -0,0 +1,89 @@
+class ApiClient {
```
jamiecarterHO: is it reasonable to call `ApiClient` `HTTPClient` or similar - I see it is being imported like

```javascript
const ApiClient = require('./httpClient');
const httpClient = new ApiClient({ baseUrl: BASE_URL });
```

Nitpick, but the rename feels awkward to me.

src/common/services/garApi.js
```diff
@@ -4,17 +4,128 @@ const endpoints = require('../config/endpoints');
...
+      return responseErrorMessage;
+    }
+  }
+
+  async getGarCheckinProgress(garId) {
+    const endpoint = `/gar/${garId}/progress`;
+    logger.debug(`Calling get GAR checkin progress endpoint ${endpoint}`);
+
+    try {
+      const { progress } = this.client.get(endpoint);
```
jamiecarterHO: Probably need to `await` this request

jamiecarterHO (changes requested):
Approach looking good and will scale well with the other APIs.

One missing `await` I think could cause issues

Just wondering about whether we can fill it out to all garAPI requests...plus a few nitpicks around naming

src/app/garfile/view/get.controller.js
ndifrekeekottHO: No it isn't I haven't re-written all of the garApi. Just updated a few for POC to see if it works. Writing new tickets that will focus on each section like Gar Upload or Organisation management. This would have made the ticket massive.

src/common/services/dataAccessApi.js
ndifrekeekottHO: Yes. So all backend calls will then be `dataAccess.{service}.{method}`

src/common/services/garApi.js
ndifrekeekottHO: This will be replaced over time. Changing this now will require me changing all calls to `exports.get` which will require refactoring a lot of areas. 

idea was to approach migration slowly as we tackle each module one at a time till we can get rid of `exports.get`.

src/test/garfile/view/post.controller.test.js
ndifrekeekottHO: I put them back

src/test/garfile/view/post.controller.test.js
```diff
@@ -133,432 +135,435 @@ describe('GAR view post controller', () => {
...
-    const cookie = new CookieModel(req);
-    cookie.setGarId('GAR-ID-EXAMPLE-1');
-
-    garApiGetStub.resolves();
-    garApiGetPeopleStub.resolves();
-    garApiGetSupportingDocsStub.rejects('garApi.getSupportingDocs Example Reject');
-    const callController = async () => {
-      await controller(req, res);
-    };
+    it('should render with an error if one of the api calls rejects', () => {
```
jamiecarterHO: This looks weirdly nested...maybe these tests won't run?

src/test/garfile/view/post.controller.test.js
```diff
@@ -43,9 +45,9 @@ describe('GAR view post controller', () => {
       render: sinon.spy(),
     };
 
-    garApiGetStub = sinon.stub(garApi, 'get');
-    garApiGetPeopleStub = sinon.stub(garApi, 'getPeople');
-    garApiGetSupportingDocsStub = sinon.stub(garApi, 'getSupportingDocs');
+    garApiGetStub = sinon.stub(dataAccessApi, 'garApi.get');
+    garApiGetPeopleStub = sinon.stub(dataAccessApi, 'garApi.getPeople');
+    garApiGetSupportingDocsStub = sinon.stub(dataAccessApi, 'garApi.getSupportingDocs');
```
jamiecarterHO: maybe `garApi` needs to be in the first argument as the second expects a method name as string

src/common/services/garApi.js
```diff
@@ -4,17 +4,129 @@ const endpoints = require('../config/endpoints');
...
-  const responseErrorMessage = JSON.stringify({
-    statusCode: _response.statusCode,
-    statusMessage: _response.statusMessage,
-    body,
-  });
-
-  return responseErrorMessage;
+function getResponseErrorMessage(err) {
+  return {
+    statusCode: err.statusCode,
```
jamiecarterHO: `status` is thrown out of the new request method instead of `statusCode`. This will be undefined for errors if they come from the new API

src/app/garfile/view/get.controller.js
```diff
@@ -66,70 +69,69 @@ module.exports = async (req, res) => {
...
-          return;
-        }
-        cookie.setGarId(parsedGar.garId);
-        cookie.setGarStatus(parsedGar.status.name);
-        logger.info(`Retrieved GAR id: ${parsedGar.garId}`);
+    try {
+      const parsedGar = garDetails;
+      const parsedPeople = garPeople;
+      const supportingDocuments = garDocs;
+      console.log('parsed gar', parsedGar);
```
jamiecarterHO: remember to remove when done testing :)

src/app/garfile/view/get.controller.js
```diff
@@ -66,70 +69,69 @@ module.exports = async (req, res) => {
...
+        res.render('app/garfile/amg/checkin/resubmit', renderContext);
+      } else {
+        logger.info(`Rendering GAR review page`);
         res.render('app/garfile/view/index', renderContext);
-      });
+      }
+    } catch (err) {
+      logger.error('Failed to get GAR information');
+      logger.error(err);
+      console.log(err);
```
jamiecarterHO: remember to remove when done testing :)

src/app/garfile/view/get.controller.js
```diff
@@ -66,70 +69,69 @@ module.exports = async (req, res) => {
...
+      renderContext.errors = [{ message: 'Failed to get GAR information' }];
+      res.render('app/garfile/view/index', renderContext);
+    }
   } catch (err) {
+    console.log(err);
```
jamiecarterHO: and here...

src/app/garfile/view/get.controller.js
ndifrekeekottHO: We running old and new side by side in some calls.

src/common/services/garApi.js
ndifrekeekottHO: @jamiecarterHO  need to understand what you mean. Should I add a `status: ...` field ?

jamiecarterHO: I think so yes...the error thrown out of the new `request` method in `httpClient.js` has a `status` param, but when creating the error message here we are looking for `statusCode` (which would work for the old API) but be `undefined` in the new one.

I think its either add or replace at the moment...not sure if both fields need to be present to surface a status code for every request that requires it.

May also have misread it - no worries if I have...

ndifrekeekottHO: I pulled this over from old code. I may need a call to understand clearly.

jamiecarterHO (approved):
Thanks for making those updates!

PR #852, opened 2026-05-28
NMSW-4856 fixed tabs pagination accessibitity issue

SaritaMore7:
fixed the issue of showing all tabbed content on a single page on a mobile view.

**Definition of Done**


src/app/home/get.controller.js
```diff
@@ -2,8 +2,10 @@ const logger = require('../../common/utils/logger')(__filename);
 const CookieModel = require('../../common/models/Cookie.class');
 const tokenApi = require('../../common/services/tokenApi');
 const garApi = require('../../common/services/garApi');
-const PER_PAGE = 10;
+
 const PAGE_ONE = 1;
+const PER_PAGE = 10;
```
ndifrekeekottHO, line 7: :nonblocking:

`PAGE_SIZE or ITEMS_PER_PAGE.`

src/app/home/get.controller.js
```diff
@@ -16,31 +18,61 @@ module.exports = (req, res) => {
   // Delete any GAR stored in the cookie session
   cookie.session.gar = null;
 
-  const statusTab = req.query?.status || 'Draft';
+  const statusTab = req.query?.status || 'Draft,Submitted,Cancelled';
```
ndifrekeekottHO: ❓ 

1. Can the parameter in status be multiple values? 
2. Can the order also be changed? - Example `Draft,Submited,Cancelled || Submitted, Cancelled,Draft`. If the other changes will this fail?

src/app/home/get.controller.js
```diff
@@ -16,31 +18,61 @@ module.exports = (req, res) => {
...
       delete req.session.successHeader;
       delete req.session.successMsg;
 
       garApi
         .getGars(userId, role, pageObj, orgId)
         .then(async (apiResponse) => {
           const garList = JSON.parse(apiResponse).items;
+
+          if (Object.keys(req.query).length === 0) {
+            req.session.garList = garList;
```
ndifrekeekottHO: :question 

Why do we need this list in session? 

The default should be - Show all until we have a filter in the query and then filter.

src/app/home/get.controller.js
```diff
@@ -16,31 +18,61 @@ module.exports = (req, res) => {
...
+          const submittedGarsList = req.session.garList.filter((gar) => {
+            if (gar.status.name === 'Submitted') {
+              return gar;
+            }
+          });
+          const cancelledGarsList = req.session.garList.filter((gar) => {
+            if (gar.status.name === 'Cancelled') {
+              return gar;
+            }
+          });
```
ndifrekeekottHO: Note: I am using session reference even though I don't get why the data is in session. But to simply the code. 

```suggestion
          const {
               Draft: draftGarsList, 
               Submitted: submittedGarsList, 
               Cancelled: cancelledGarsList
               } = Object.groupBy(req.session.garList, (item) => item.status.name)
```

The Object.groupBy will give you `{Draft: [...], Cancelled: [... ], Submitted: [...]}`

src/common/models/Cookie.class.js
```diff
@@ -88,6 +88,9 @@ class Cookie {
     if (this.session.redirectUrl == null) {
       this.session.redirectUrl = '';
     }
+    if (this.session.garList == null) {
```
ndifrekeekottHO: I would like to understand why we keeping the list in Session. Can't we derive the list we need at run time?

ndifrekeekottHO (changes requested):

src/common/models/Cookie.class.js
SaritaMore7: I have used the session here to store the garlist with all three statuses during the initial page load. This ensures that when pagination starts for a particular status, it does not overwrite the values from other tab pages that were loaded initially.
If we use runtime instead, I will not be able to retain and display those gars on the other tabs by calling endpoint only once to fetch the data.

ho-henry-senior (approved):

jamiecarterHO (approved):
Great work!

ndifrekeekottHO (approved):

PR #855, opened 2026-06-04
fixed the js issue on resp person page

SaritaMore7:
fixed the js issue on resp person page - The HTML is not loaded when the 'Responsible Person' tab is selected. so added a condition to check whether the target element exists before attaching the event listener to the selected element.

**Definition of Done**

### DIFF
--- a/src/public/javascripts/includes/resperson.js
+++ b/src/public/javascripts/includes/resperson.js
@@ -42,7 +42,9 @@ $(document).ready(function () {
     });
   });
 
-  accessibleAutocomplete.enhanceSelectElement({
-    selectElement: document.querySelector('#responsibleCountry'),
-  });
+  if (document.getElementById('#responsibleCountry')) {
+    accessibleAutocomplete.enhanceSelectElement({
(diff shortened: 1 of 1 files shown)


src/public/javascripts/includes/resperson.js
```diff
@@ -42,7 +42,9 @@ $(document).ready(function () {
     });
   });
 
-  accessibleAutocomplete.enhanceSelectElement({
-    selectElement: document.querySelector('#responsibleCountry'),
-  });
+  if (document.getElementById('#responsibleCountry')) {
```
ho-henry-senior, line 45: does this fix a console error when #responsibleCountry does not exist?

SaritaMore7: yes

ho-henry-senior (approved):

src/public/javascripts/includes/resperson.js
SaritaMore7: image

After fix:

image

PR #859, opened 2026-06-11
fixed the departure date validation issue  on review page - v5 changes

SaritaMore7:
When you click Submit GAR for passenger check and if departure date is less than 2 hours,  we used to have an warning error message with the content. It was missing in the latest changes. fixed this issue in this PR.

**Definition of Done**

### DIFF
--- a/src/common/templates/check_your_answers/macros.njk
+++ b/src/common/templates/check_your_answers/macros.njk
@@ -22,7 +22,9 @@
 {% endmacro %}
 
 {% macro transitionDetails(prefix, garfile, showChangeLinks, __) %}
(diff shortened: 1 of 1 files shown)


src/common/templates/check_your_answers/macros.njk
```diff
@@ -22,7 +22,9 @@
 {% endmacro %}
 
 {% macro transitionDetails(prefix, garfile, showChangeLinks, __) %}
-  <div id="{{ prefix ~ 'Date' }}"></div>
```
ho-henry-senior, line 25: why is this line removed?

ho-henry-senior (approved):

PR #866, opened 2026-07-01
NMSW 4655 update incorrect error message

TemitopeAyokuHO:
[NMSW 4655](https://jira.bics-collaboration.homeoffice.gov.uk/browse/NMSW-4655) - Uploading more than the maximum number of supporting docs shows incorrect error message

**Definition of Done**

### DIFF
--- a/src/app/garfile/supportingdocuments/get.controller.js
+++ b/src/app/garfile/supportingdocuments/get.controller.js
@@ -38,6 +38,10 @@ module.exports = (req, res) => {
           error = [{ identifier: 'file', message: 'No file selected for upload' }];
           break;
         case 'limit':
+          error = [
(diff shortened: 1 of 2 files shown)


src/app/garfile/supportingdocuments/get.controller.js
```diff
@@ -38,14 +38,8 @@ module.exports = (req, res) => {
           error = [{ identifier: 'file', message: 'No file selected for upload' }];
           break;
         case 'limit':
-          error = [
```
ndifrekeekottHO, line 41: @TemitopeAyokuHO  Don't we still need the error message that says files cannot be more than 7.5MB? 

Ticket seemed to focus on number of files uploaded error message and not file size as that is still in place.

ndifrekeekottHO (changes requested):

ndifrekeekottHO (approved):

PR #870, opened 2026-07-08
Added Downscaler dynamic annotations to scale down notprod environments only!

b-pb:

**Definition of Done**

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -119,7 +119,6 @@ steps:
     commands:
       - source ./ops/dev_deploy_vars.sh
       - source ./ops/deploy.sh
-      - source ./ops/notprod_values
     environment:
       INGRESS_CLASS: nginx-internal
       INSECURE_SKIP_TLS_VERIFY: true
@@ -157,7 +156,6 @@ steps:
     commands:
       - source ./ops/test_deploy_vars.sh
       - source ./ops/test_deploy.sh
-      - source ./ops/notprod_values
     environment:
       INGRESS_CLASS: nginx-internal
       INSECURE_SKIP_TLS_VERIFY: true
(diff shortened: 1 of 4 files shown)


.drone.yml
```diff
@@ -119,6 +119,7 @@ steps:
     commands:
       - source ./ops/dev_deploy_vars.sh
       - source ./ops/deploy.sh
+      - source ./ops/notprod_values
```
jamiecarterHO: I believe these values are already be available from being exported via `dev_deploy_vars.sh` - same for all similar statements

ops/prod_values
```diff
@@ -1,2 +1,3 @@
 #values for prod deployments
-export NUM_REPLICAS=3 
\ No newline at end of file
+export NUM_REPLICAS=3 
+export DOWN_SCALER_ANNOTATIONS="environment: 'egar-production'"
```
jamiecarterHO: if this is to avoid a non-existent variable issue for uptime in prod then you should probably either use an uptime that represents 24/7 or make the annotations in the deployment environment specific e.g.

```bash
{{ if ne .KUBE_NAMESPACE "egar-production" }}
annotations:
  downscaler/uptime: {{ .DOWN_SCALER_ANNOTATIONS }}
{{ end }}
```

b-pb (commented):
Hi @jamiecarterHO I had followed your advice and put the conditions in the main deployment manifest and removed any references from the prod_values

b-pb (commented):
Please see latest 'Files changed'

ops/prod_values
```diff
@@ -1,2 +1,2 @@
 #values for prod deployments
-export NUM_REPLICAS=3 
\ No newline at end of file
+export NUM_REPLICAS=3
```
jamiecarterHO: I'm not sure if it will matter, but might be best to clear the extra whitespace added on the end of this line

.drone.yml
jamiecarterHO: I mean that each of the new lines in this file that read `- source ./ops/notprod_values` or `- source ./ops/prod_values` can probably be removed as `- source ./ops/dev_deploy_vars.sh` has inside `source ./ops/notprod_values`

jamiecarterHO (approved):

PR #876, opened 2026-07-23
feat: Add exclusion list for csrf generation

ndifrekeekottHO:

We want to prevent generating CSRF for every request route

**Definition of Done**

### DIFF
--- a/src/common/config/csrfExclusionList.js
+++ b/src/common/config/csrfExclusionList.js
@@ -1,11 +1,14 @@
 module.exports = {
   CSRF_EXCLUSION_LIST: [
     '/help',
+    '/login',
+    '/user/logout',
     '/user/details',
     '/public/.*',
     '/assets/.*',
     '/javascripts/.*',
     '/stylesheets/.*',
     '/',
+    '/error/.*',
   ],
 };
--- a/src/common/config/csrfExclusionList.js
+++ b/src/common/config/csrfExclusionList.js
@@ -1,0 +1,11 @@
+module.exports = {
+  CSRF_EXCLUSION_LIST: [
+    '/help',
+    '/user/details',
+    '/public/.*',
+    '/assets/.*',
+    '/javascripts/.*',
+    '/stylesheets/.*',
+    '/',
+  ],
+};
--- a/src/common/middleware/csrfcheck.js
+++ b/src/common/middleware/csrfcheck.js
@@ -5,11 +5,11 @@ if (process.env.COOKIE_SECURE_FLAG === 'true') {
   secureFlag = true;
 }
 module.exports = (_req, _res, next) => {
-  csurf({
-    cookie: {
-      httpOnly: true,
-      secure: secureFlag,
-    },
-  });
+  // csurf({
+  //   cookie: {
+  //     httpOnly: true,
+  //     secure: secureFlag,
+  //   },
+  // });
   next();
 };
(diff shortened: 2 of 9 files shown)


src/server.js
```diff
@@ -125,30 +124,6 @@ function initialiseGlobalMiddleware(app) {
     })
   );
 
-  app.use((req, res, next) => {
-    res.locals.asset_path = '/public/';
-    noCache(res);
-    const token = req.csrfToken();
-    res.locals._csrf = token;
```
ndifrekeekottHO, line 132: TODO: 

- Confirm the use of `res.locals.assert_path` and `res.locals._csrf`.

ndifrekeekottHO: @jamiecarterHO  confirmed that res.locals is used to set variable that can be access from templates. 

Though it seems the use isn't consistent - example, we already setting `csrfToken` in line 145 below. Maybe be best to leave as is to avoid risk elsewhere in the codebase.

src/test/common/middleware/csrfcheck.test.js
```diff
@@ -11,49 +8,60 @@ describe('CSRF Check Middleware', () => {
...
+    const requestPaths = ['/home', '/organisation/assignrole', '/unknown_path'];
+
+    sinon.spy(req, 'csrfToken');
+
+    for (const path of requestPaths) {
+      req.path = path;
+
+      await proxiedMiddleware(req, res, next);
+
+      expect(req.csrfToken).to.have.be.called;
```
jamiecarterHO:
```suggestion
      expect(req.csrfToken).to.have.been.called;
```

src/server.js
```diff
@@ -125,30 +123,6 @@ function initialiseGlobalMiddleware(app) {
...
-    const _render = res.render;
-    res.render = function (view, options, fn) {
-      _.extend(options, { csrfToken: token });
-      _render.call(this, view, options, fn);
-    };
-
-    next();
-  });
-
   logger.info('Set CSRF Token');
```
jamiecarterHO: This line has moved and is less accurate here as it may not always be set

src/common/config/csrfExclusionList.js
jamiecarterHO, line 1: Can you document this file in the README - feels like there should be an up front indication that this is how we are managing CSRF use

src/server.js
jamiecarterHO: it doesn't look like `locals.asset_path` is accessed anywhere and it doesn't seem to be a csurf config. 

`locals._csrf` seems to be unused while `csrfToken` from the render extension is. 

I wonder if we can just chop them out?

src/common/middleware/csrfcheck.js
```diff
@@ -1,15 +1,33 @@
-const csurf = require('csurf');
+const logger = require('../../common/utils/logger')(__filename);
+const noCache = require('../utils/no-cache');
+const { canSkipCsrfGeneration } = require('../utils/utils');
+const _ = require('lodash');
 
 let secureFlag = false;
 if (process.env.COOKIE_SECURE_FLAG === 'true') {
   secureFlag = true;
 }
```
jamiecarterHO: `secureFlag` is not used here anymore

jamiecarterHO (commented):
Looks good overall...no absolute blockers in comments (except maybe the test call typo)

Aside from other comments we could probably remove the extra calls to `csrfCheck` where dupe set at page router level e.g.
  - src/app/organisation/assignrole/index.js
  - src/app/verify/organisationinvite/index.js
  - src/app/verify/registeruser/index.js
  - maybe others?

jamiecarterHO (approved):
Thanks for updates

PR #877, opened 2026-07-28
NMSW 5012 Harden id token storage

TemitopeAyokuHO:
[NMSW-5012](https://jira.bics-collaboration.homeoffice.gov.uk/browse/NMSW-5012) - Harden-id-token-storage

**Definition of Done**


src/app/user/login/get.controller.js
```diff
@@ -68,7 +68,7 @@ const sendAdminUpdateEmail = (userObj) => {
  * @param {Object} cookie - Cookie model instance
  * @returns {Promise<Object>} - User authentication result
  */
-const handleUserAuthentication = (req, res, userInfo, cookie) => {
+const handleUserAuthentication = (req, res, userInfo, cookie, idToken) => {
```
ho-henry-senior: if `id_token` is available from `req.session` why have you added it as an extra argument?

ho-henry-senior (approved):
Thanks Temi, there are a couple of other Copilot suggestions, but I think this is good to go.

PR #882, opened 2026-08-03
Nmsw 5010 regenerate session on login

TemitopeAyokuHO:
[NMSW 5010](https://jira.bics-collaboration.homeoffice.gov.uk/browse/NMSW-5010) - Regenerate-session-on-login

- added a session regeneration function in the login controller, this will be applied after user is authenticated 
- added tests for ID rotation plus session-state migration.

**Definition of Done**


src/test/user/login/session-regeneration.test.js
```diff
@@ -0,0 +1,124 @@
...
+
+    sinon.stub(userApi, 'getDetails').resolves({
+      organisation: { organisationId: 'ORG-1', organisationName: 'Org Name' },
+    });
+
+    await controller(req, res);
+
+    expect(regenerateStub).to.have.been.calledOnce;
+    expect(req.session.preAuthMarker).to.equal('keep-me');
+    console.log('req.session after regeneration:', req.session);
```
jamiecarterHO: Can take out this log now

src/app/user/login/get.controller.js
```diff
@@ -68,7 +101,7 @@ const sendAdminUpdateEmail = (userObj) => {
  * @param {Object} cookie - Cookie model instance
```
jamiecarterHO, line 69: remove this param from the docs if taking it out of the function

src/app/user/login/get.controller.js
```diff
@@ -270,3 +310,5 @@ function redirectErrorPage(req, res, errorPage) {
 
   return getOneLoginLogoutUrl(req, sessionToken, req.cookies.state);
 }
+
+module.exports.sessionRegenerateForAuthenticatedUser = sessionRegenerateForAuthenticatedUser;
```
jamiecarterHO: Might be better to create a new util file for this to import from

src/app/user/login/get.controller.js
```diff
@@ -199,15 +231,23 @@ module.exports = async (req, res) => {
           }
           return oneLoginApi.getUserInfoFromOneLogin(access_token);
         })
-        .then((userInfo) => {
+        .then(async (userInfo) => {
```
jamiecarterHO: don't need the `async` keyword if `await` is not being used

src/app/user/login/get.controller.js
```diff
@@ -199,15 +231,23 @@ module.exports = async (req, res) => {
...
+            .then((isRegenerated) => {
+              if (!isRegenerated) {
+                return { redirect: redirectErrorPage(req, res, 'service-error') };
+              }
+
+              return handleUserAuthentication(req, res, userInfo);
+            })
             .then(({ redirect }) => {
+              const cookie = new CookieModel(req);
               const redirectUrl = cookie.getRedirectUrl();
```
jamiecarterHO, line 219: It looks like whatever `redirectUrl` is from `cookie` would override the `{ redirect }` passed to this part of the the promise chain

src/app/user/login/get.controller.js
```diff
@@ -26,6 +26,39 @@ const USER_STATES = {
...
+ * @returns {Promise<boolean>} - True if regeneration succeeds
+ */
+
+const sessionRegenerateForAuthenticatedUser = (req) => {
+  if (!req?.session || typeof req.session.regenerate !== 'function') {
+    logger.error('Session regeneration unavailable during login flow.');
+    return Promise.resolve(false);
+  }
+
+  const sessionDataToMigrate = Object.keys(req.session).reduce((acc, key) => {
```
jamiecarterHO: May be worth creating a deep clone of the session to avoid any accidental reference links to the older object

We could also be more strict about an allowlist for what is transitioned session to session - what do we _need_ to keep? 

It would also be better to preserve keys, build a new session and set back what we need in one go

e.g.

```javascript
const sessionRegenerateForAuthenticatedUser = (req) => {
  if (!req?.session || typeof req.session.regenerate !== 'function') {
    logger.error('Session regeneration unavailable during login flow.');
    return Promise.resolve(false);
  }

  // only keep what we need to absolute minimum
  const keysToPreserve = [
    'redirectUrl',
    'currentPage',
  (anything else)
  ];

  const preservedValues = keysToPreserve.reduce((acc, key) => {
    if (req.session[key] !== undefined && typeof req.session[key] !== 'function') {
      acc[key] = structuredClone(req.session[key]);
    }
    return acc;
  }, {});

  return new Promise((resolve) => {
    req.session.regenerate((error) => {
      if (error) {
        logger.error(`Failed to regenerate session during login: ${error}`);
        return resolve(false);
      }

      // Rebuild default session shape on the fresh session

      // Restore only the allowed values
      Object.assign(cookie.session, preservedValues);

      return resolve(true);
    });
  });
};
```

jamiecarterHO (changes requested):
I think it may be wise to add a session regenerate in src/app/user/onelogin/post.controller.js where the user completes the registration flow around line 120

src/test/user/login/session-regeneration.test.js
jamiecarterHO, line 1: These are great, would be good to have a negative test for the scenario when sessionRegenerateForAuthenticatedUser resolves false or throws an error

src/common/utils/session_generator.js
```diff
@@ -0,0 +1,40 @@
...
+ * Regenerates the session identifier and migrates non-cookie, non-function state.
+ * This prevents a pre-auth session identifier from being reused after login.
+ */
+
+const sessionRegenerateForAuthenticatedUser = (req) => {
+  if (!req?.session || typeof req.session.regenerate !== 'function') {
+    logger.error('Session regeneration unavailable during login flow.');
+    return Promise.resolve(false);
+  }
+  const keysToPreserve = ['redirectUrl', 'currentUser', 'preAuthMarker'];
```
jamiecarterHO: Maybe have a check that there isn't anything else beyond these to migrate over if you haven't already - I'm not sure what is required myself. 

You could also use `redirectUrl` as a check instead of adding the `preAuthMarker` for test purposes only

jamiecarterHO (commented):
Make sure to add the new session regen into the part of src/app/user/onelogin/post.controller.js where the user completes their registration sign-in flow

src/common/utils/session_generator.js
TemitopeAyokuHO: I believe we should only these keys
- redirectUrl
- currentUser
- preAuthMarker

And we expect every other keys gets a fresh session context unless re-populated after auth. i have replaced preAuthMarker with redirectUrl in test

jamiecarterHO: can take out `preAuthMarker` from this array now

jamiecarterHO (changes requested):
@TemitopeAyokuHO I believe we also need to add the session regenerate to the part of the flow when a user is signing up and they complete One Login authentication. This is in `src/app/user/onelogin/post.controller.js` around line 120

